PR #4, opened 2025-12-07
feat(output): Display more device-friendly information when outputting device list

bnqzzdf:
Added display of target device name

Display device output name and configuration information

Output product ID in hexadecimal format

Display protocol version number

Indicate whether multi-port support is enabled

Display synchronization status and its detailed parameters


cometdom:
Hi again @bnqzzdf,

And thanks a lot for these new adds.
As usual they are very useful.
Would it be indiscreet to ask your name?
It would make our exchanges more user-friendly.

Best regards,
Dominique.

PR #9, opened 2025-12-11
Feature/divers

herisson-88:
Bonjour Dominique,

Tu peux tester ces évolutions, je ne serai pas vexé si tu refuse tout ou partie ;-)

# Pull Request: Stability and Thread-Safety Improvements

## Summary

This PR introduces several improvements focused on stability, performance, and thread-safety for the DirettaRendererUPnP project.

### 1. Optimize FFmpeg settings for faster streaming startup
- Add `probesize` (1MB) and `analyzeduration` (1.5s) for faster initial buffering
- Increase network buffer from 32KB to 512KB for better stability on slow connections

### 2. Disable gapless playback on format changes
- Detect format changes (sample rate, bit depth, channels, DSD vs PCM) between tracks
- Force clean stop/start sequence instead of gapless transition to prevent audio artifacts

### 3. Add anti-pop protection with silence padding
- Add `sendSilence()` function to pre-fill buffer with format-appropriate silence (0x69 for DSD, zeros for PCM)
- Apply anti-pop at playback start and after format changes
- Note: `drainBuffer()` not added as drain functionality already exists using `buffer_empty()`

### 4. Add thread-safe track change mechanism
- Add dedicated `m_pendingMutex` to protect string writes/reads (`std::string` is not atomic)
- Use `memory_order_acquire/release` for ARM compatibility
- `setCurrentURI()` / `setNextURI()` now queue changes via pending mechanism when playing
- `process()` applies pending changes safely in audio thread context

### 5. Fix use-after-free in preload thread
- Replace `std::thread().detach()` with managed joinable thread
- Add `m_preloadThread` and `m_preloadRunning` to track preload state
- Add `waitForPreloadThread()` helper called in `stop()` and destructor
- Prevents crash when `AudioEngine` is destroyed while preload is running

### 6. Fix race condition on preloadNextTrack()
- Add `m_preloadRunning` check in `process()` before calling `preloadNextTrack()`
- Prevents simultaneous preload from `play()` thread and `process()` audio thread

## Files Modified

- `src/AudioEngine.cpp` - FFmpeg optimization, gapless detection, thread-safety, preload fixes
- `src/AudioEngine.h` - New members for pending mechanism and preload thread management
- `src/DirettaOutput.cpp` - `sendSilence()` function, anti-pop in `changeFormat()`
- `src/DirettaOutput.h` - `sendSilence()` declaration
- `src/DirettaRenderer.cpp` - Anti-pop at playback start

## Testing

- [ ] Playback start without pops/clicks
- [ ] Format change between tracks (e.g., 44.1kHz to 96kHz)
- [ ] Gapless playback with same format tracks
- [ ] Rapid play/stop/play sequences
- [ ] Application shutdown during playback

## Compatibility

- Based on `develop-v1.0.6` branch (commit `88c2046`)
- No breaking changes to existing API

### DIFF
--- a/src/AudioEngine.cpp
+++ b/src/AudioEngine.cpp
@@ -1232,7 +1232,9 @@ bool AudioEngine::process(size_t samplesNeeded) {
     
     // Preload next track as soon as EOF flag is set (for gapless)
     // Check AFTER readSamples() because EOF flag is set during the read
-    if (!m_nextDecoder && !m_nextURI.empty() && m_currentDecoder->isEOF()) {
+    // Also check m_preloadRunning to avoid race with preload thread from play()
+    if (!m_nextDecoder && !m_nextURI.empty() && m_currentDecoder->isEOF()
+        && !m_preloadRunning.load(std::memory_order_acquire)) {
         std::cout << "[AudioEngine] EOF detected, preloading next track for gapless..." << std::endl;
         preloadNextTrack();
     }


herisson-88:
Et ce n'est pas grave du tout si ce n'est pas mergé si tu préfère reprendre comme tu veux

PR #14, opened 2025-12-12
Feature/multi thread

herisson-88:
Fix thread-safety issues causing silent playback and crashes

  Problem

  When skipping tracks rapidly, the renderer experienced:
  - Silent playback (no audio output)
  - Segmentation faults
  - Race conditions visible in logs (interleaved messages)

  Root Causes

  1. Race condition in setNextURI() (AudioEngine)
    - UPnP thread writes to m_nextURI/m_nextDecoder while audio thread reads them
    - Cross-thread string manipulation without synchronization
  2. Use-after-free in preload thread (AudioEngine)
    - Detached thread continues running after AudioEngine destruction
    - Accesses freed memory → crash
  3. Concurrent UPnP callbacks (DirettaRenderer)
    - Multiple UPnP actions (Stop, Play, SetURI) execute in parallel
    - Interleaved operations corrupt internal state

  Solution

  Commit 1: Fix thread-safety issues in AudioEngine
  - Pending mechanism for setNextURI(): UPnP thread writes to pending variables, audio thread applies    
   them safely
  - Joinable thread instead of detached: destructor waits for preload thread completion
  - Memory ordering with std::memory_order_acquire/release

  Commit 2: Serialize UPnP action callbacks
  - Added std::lock_guard<std::mutex> to all 6 UPnP callbacks
  - Ensures only one action executes at a time

  Testing

  - Rapid track skipping: ✅ No more silent playback
  - Stress test (skip 10+ tracks quickly): ✅ No crashes
  - Gapless playback: ✅ Still works correctly

### DIFF
--- a/src/DirettaRenderer.cpp
+++ b/src/DirettaRenderer.cpp
@@ -307,6 +307,7 @@ bool DirettaRenderer::start() {
 UPnPDevice::Callbacks callbacks;
 
 callbacks.onSetURI = [this](const std::string& uri, const std::string& metadata) {
+    std::lock_guard<std::mutex> lock(m_mutex);  // Serialize UPnP actions
     std::cout << "[DirettaRenderer] SetURI: " << uri << std::endl;
     
     // ⭐ Sauvegarder l'URI courante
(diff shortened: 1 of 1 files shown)


herisson-88:
C'est un plus gros morceau mais avec ça j'arrive à changer de pistes comme un parkinsonien sans crash ;-)

PR #22, opened 2025-12-16
DSD512, DSD1024 support added

norman-arch:
Hello,

dsd512 and dsd1024 playback added. Have tested on my music system, thanks.


norman-arch:
thank you, will close then.

PR #23, opened 2025-12-17
Fix race condition between audio callback and close()

herisson-88:
Une nouvelle tentative de fix des races... Tu me dis si cela améliore... (du moi chez moi pas de regression sur mconnect ou hifi cast
Problem:
The audio callback was accessing DirettaOutput while the UPnP thread called close(), causing undefined behavior and potential crashes. The stopTimeMutex was also redundant since m_mutex already serializes UPnP callbacks.

Root cause analysis:
- Thread Audio: sendAudio(m_direttaOutput) concurrent with
- Thread UPnP: m_direttaOutput->close()
- No synchronization between callback execution and resource cleanup

Solution:
1. Add m_callbackMutex + condition_variable for callback/close sync
2. Callback checks state and sets running flag atomically with mutex
3. UPnP calls stop() with mutex held, then waits for callback completion
4. RAII guard (CallbackGuard) ensures flag cleanup on any exit path
5. 5-second timeout prevents infinite blocking on bugs

Simplification:
- Remove stopTimeMutex (redundant with m_mutex serialization)
- Keep DAC stabilization delay (hardware requirement, not race fix)

Safety improvement:
- Add decoder auto-reopen in process() as defensive safety net
- Handles edge case where decoder is null while state is PLAYING

Deadlock analysis:
- No deadlock possible: try_to_lock never blocks, CV wait releases mutex
- Audio thread never acquires m_mutex, breaking potential cycles

Files modified:
- DirettaRenderer.h: Add sync primitives and waitForCallbackComplete()
- DirettaRenderer.cpp: Sync in callback, onSetURI, onStop; remove stopTimeMutex
- AudioEngine.cpp: Safety net for null decoder recovery

### DIFF
--- a/src/AudioEngine.cpp
+++ b/src/AudioEngine.cpp
@@ -1285,9 +1285,26 @@ bool AudioEngine::process(size_t samplesNeeded) {
         std::cout << "[AudioEngine] Pending next URI applied (gapless)" << std::endl;
     }
 
+    // Safety net: auto-reopen if decoder null while PLAYING
     if (!m_currentDecoder) {
-        return false;
-    }    // ... reste du code ...    // Determine output format
+        if (!m_currentURI.empty()) {
+            if (!openCurrentTrack()) {
+                std::cerr << "[AudioEngine] Failed to reopen track" << std::endl;
+                m_state = State::STOPPED;
+                if (m_trackEndCallback) {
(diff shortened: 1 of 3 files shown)


herisson-88:
je pousserai quand j'aurai un truc qui marche avec Roon

PR #65, opened 2026-04-12
Ensure the linker uses lld & Optimize compiler flags

sheviks:
# Optimize LDFLAGS for LTO and unify optimization levels to -O3

## Description

This PR introduces optimizations to the build system's flags to ensure better performance and consistency, particularly when LTO is utilized.

The changes are categorized into three parts:

### 1. Synchronize LDFLAGS with Compiler Flags for LTO
When LTO is enabled, the linker handles the final code generation. To ensure it applies the correct architecture-specific optimizations and instruction scheduling, `LDFLAGS` now includes the same `-O` and `-march` flags used in compilation. This prevents the linker from falling back to generic instructions during the Whole Program Analysis phase.

### 2. Best Practice & Safety
For non-LTO builds, adding these flags to `LDFLAGS` is idempotent and harmless. This logic follows the standard behavior of modern build systems; for example, **CMake-generated Makefiles** (as seen in projects `slim2Diretta`) consistently propagate these flags to the linker to ensure build integrity.

### 3. Unified Optimization Level (-O3)
I noticed a discrepancy where `src/fastmemcpy-avx.c` used `-O3` while most `.cpp` files used `-O2`. To maintain performance consistency across the project, I have unified these to `-O3`.

> **Note:** If the use of `-O2` for C++ files was an intentional design choice (e.g., to avoid specific compiler regressions), please feel free to ignore or revert the final commit of this PR.

Thank you!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,9 +17,9 @@ else
     CXX = g++
     CC = gcc
 endif
-CXXFLAGS = -std=c++17 -Wall -Wextra -O2 -pthread
+CXXFLAGS = -std=c++17 -Wall -Wextra -O3 -pthread
 CFLAGS = -O3 -Wall
-LDFLAGS = -O2 -pthread
+LDFLAGS = -O3 -pthread
 ifdef LLVM
     CXXFLAGS += -flto
     CFLAGS += -flto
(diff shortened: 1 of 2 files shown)


sheviks:
Please hold off on processing this PR for now. I just realized I might have missed something and would like to do a final check.

sheviks:
The issue where `clang` might use the system default linker instead of `lld` has been fixed.

Additionally, regarding commit 3edcd26, please decide whether to unify the optimization level to -O2, -O3, or retain the current setting.

Please review this PR.
Thanks again.

PR #67, opened 2026-04-20
Option to set the target network speed and duplex mode in the webui

Koala887:
Since network speed has a significant impact on sound quality, I've added an option to set the target network speed and duplex mode in the web interface.
10 Mbps is perfectly sufficient for 44.1 – 96kHz files.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1210,7 +1210,19 @@ IO_SCHED_CLASS="${IO_SCHED_CLASS:-realtime}"
 IO_SCHED_PRIORITY="${IO_SCHED_PRIORITY:-0}"
 RT_PRIORITY="${RT_PRIORITY:-50}"
 
+# Advanced network config
+TARGET_INTERFACE="${TARGET_INTERFACE:-}"
+TARGET_SPEED="${TARGET_SPEED:-100}"
+TARGET_DUPLEX="${TARGET_DUPLEX:-full}"
+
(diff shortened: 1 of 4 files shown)


cometdom:
Hi @Koala887 

Thanks again for the contribution — your PR #67 is merged into dev-2.4.0, so the option will land in the next v2.4.0 release.

Just so you know what to expect when you pull the branch: I added a small follow-up commit on top of yours (3c3a47a) with a few hardening tweaks, none of which change the behavior you implemented:

Shell-quoted the variables passed to ethtool in both start-renderer.sh and install.sh (defensive against odd characters in interface names entered via the web UI).
Wrapped the ethtool call in a command -v ethtool check so the launcher prints a clear warning instead of failing silently when the package isn't installed.
Added ethtool to the base dependency lists (dnf/apt/pacman) in install.sh.
Added a bandwidth-vs-format warning to the .conf comment and the web UI description, since 100 Mbit underruns DSD512+ (~45 Mbps raw) and 10 Mbit only covers up to ~96 kHz PCM — I wanted to make sure an audiophile reflexively picking 10 Mbit doesn't lose hi-res playback without an obvious reason.
The opt-in path you wrote is untouched. Thanks for spotting this gap and getting it implemented — link tuning is one of those audiophile-niche knobs that's a real pain to find when you need it, and now it's in the box.

Dominique

PR #68, opened 2026-05-05
CPU affinity for the DirettaRenderer Audio thread (decode) + Bugfix NIC IRQ affinity

Koala887:
Added a option to change the CPU affinity of the DirettaRenderer audio thread. Receiving and decoding the stream should run on a dedicated core, separate from the UPnP, position and logging threads.
Removed #ProtectKernelTunables=true in diretta-renderer.service so that the NIC IRQ affinity can be changed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1193,11 +1193,14 @@ setup_systemd_service() {
 set -e
 
 # Default values (can be overridden by config file)
+# v2.1.10: Aligned variable names with CLI (KEY → --key mapping)
+# Old names (RENDERER_NAME, NETWORK_INTERFACE, MTU_OVERRIDE) still supported as fallback
 TARGET="${TARGET:-1}"
 PORT="${PORT:-4005}"
 NAME="${NAME:-${RENDERER_NAME:-}}"
 GAPLESS="${GAPLESS:-}"
 VERBOSE="${VERBOSE:-}"
+MINIMAL_UPNP="${MINIMAL_UPNP:-}"
 INTERFACE="${INTERFACE:-${NETWORK_INTERFACE:-}}"
 THREAD_MODE="${THREAD_MODE:-}"
 CYCLE_TIME="${CYCLE_TIME:-}"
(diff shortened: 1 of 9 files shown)


cometdom:
Hi @Koala887, 
I will check this PR as soon as possible.
Please note that if you are using DRUP in GentooPlayer or Audiolinux, options like IRQ affinity are no longer supported in these distributions.

cometdom:
Hi @Koala887 

Thanks again for this PR — both contributions are excellent and have just been merged into main as commit dae8273, with v2.4.2 prepared and the CHANGELOG / README / docs updated to credit you.

A few words on each change:

--cpu-decode — really nice extension of the original CPU affinity work from PR #67. The three-tier model (--cpu-audio for the Diretta SDK worker, --cpu-decode for the renderer audio thread, --cpu-other for everything else) maps cleanly to what an audiophile setup wants: maximum isolation of the critical hot path, separation of the decode work, and a free hand for kernel/IRQ activity. The fact that you also raise the audio thread to SCHED_FIFO when a dedicated core is available is the right call — predictable scheduling on a reserved core is exactly when RT priority is safe and useful. Setting --cpu-decode empty preserves the v2.4.1 fallback behaviour, so no existing user is impacted.

ProtectKernelTunables=true fix — well caught, and honestly that one was a regression I introduced in v2.4.0 without realizing. When I added IRQ_INTERFACE / IRQ_CPUS writing to /proc/irq/N/smp_affinity_list from the wrapper, I didn't notice that the systemd unit was sandboxing it away. Your fix restores the IRQ affinity feature to actually working — sorry for the oversight, and thanks for spotting it.

I made a few minor post-merge tweaks (commit 827ea1f): cleaned up some mixed tabs/spaces in audioThreadFunc(), renamed the inner cores variables to decodeCores / otherCores for readability, version bump to 2.4.2, and the CHANGELOG / README / docs/CLAUDE.md updates with proper credit to you. Nothing functional, just polish.

Both changes will ship in the upcoming v2.4.2 release, which Dominique is preparing now. If you want to mirror the same --cpu-decode granularity into slim2diretta at some point, that would be welcome there too — but no pressure, and I can also port it myself when there's a quieter moment.

Thanks again for the careful contribution and the clear PR description.

Cheers,
Dominique

PR #72, opened 2026-05-16
fix: clean shutdown on corrupt PCM packet from radio stream (v2.4.5)

hoorna:
## Problem

A corrupt packet mid-stream (e.g. from an internet radio source) caused `avcodec_receive_frame()` to return an error **after** some samples had already been decoded in the same `readSamples()` call. Because the error check was guarded by `samplesRead == 0`, it was silently skipped — leaving the decoder flagged as failed while the renderer kept running, producing silence and ignoring all subsequent UPnP commands (**zombie state**).

## Fix

**`src/AudioEngine.cpp`**
- Decode-error check moved before the `samplesRead == 0` guard so it fires even on partial reads
- On detection: preload thread is joined, next-track state (`m_nextDecoder`, `m_nextURI`, `m_nextMetadata`, `m_formatChangePending`) is cleared
- `m_state` set to `STOPPED` before firing `m_trackEndCallback()` — mirrors the normal EOF teardown path, ensuring the UPnP controller (including Roon) sees the correct state before transitioning

**`src/AudioEngine.h`**
- `hasDecodeError()` doc comment tightened to reflect the "distinct from EOF" distinction

## Test plan

- [ ] Play a radio stream that emits a corrupt PCM packet — renderer should stop cleanly and return to idle rather than hanging
- [ ] Normal EOF playback (local file, playlist end) is unaffected
- [ ] Preloaded next track is discarded cleanly on error, no dangling thread


cometdom:
Hi @hoorna, thanks a lot for this — excellent catch and a very clean fix. 🙏

Merged to `main` (squashed into a single commit for a tidy history). What I particularly appreciated reviewing it: you didn't invent a new teardown path, you mirrored the existing EOF one (`m_state = STOPPED` → `m_trackEndCallback()` under `m_mutex`, the `unlock → waitForPreloadThread → lock` dance). That's exactly the right instinct on this code — it means the change carries no new deadlock surface, it reuses a path that's been in production for a long time.

Two tiny **cosmetic** tweaks I applied on top after merge, just so you know (no code change):

1. `(PR #1)` → `(PR #72 by hoorna/Alfred)` in CHANGELOG / docs/CLAUDE.md — it was your PR #1 on your fork, #72 here. Wanted the upstream history to credit you correctly.
2. Reworded "matching the normal EOF teardown path *exactly*" → "mirrors the EOF teardown's state-then-callback ordering". The corrupt-packet path intentionally **skips** the ~25 ms end-of-track drain (the 5 `m_silenceCount` cycles) the EOF path does — which is correct (a corrupt packet isn't a clean end), I just wanted the docs to be precise about that intentional difference.

Heads-up on timing: I'm **holding the v2.4.5 release** to batch it with upcoming changes (v2.4.4 just shipped), so the tag won't appear immediately — the fix is on `main` and will go out with the next release.

Two non-blocking thoughts for possible future work, purely for discussion — not asks:

- **Clean stop vs. skip-and-continue on radio.** Right now a corrupt packet stops playback cleanly (which is the correct fix for the zombie state). For internet radio specifically, some users might prefer the renderer to drop the bad packet and keep going (a brief glitch vs. a full stop). It's debatable — after a corrupt packet the decoder is often desynced anyway — but if you ever want to explore a "resync and resume" mode for radio sources, I'd be happy to look at it.
- **`waitForPreloadThread()` has no timeout** (it's a bare `join()`). Pre-existing, not introduced by your PR, and unlikely to bite on radio (no next track → `m_preloadRunning` is false). Just flagging it as known tech debt in case you bump into it.

Thanks again — really solid contribution. 👏

PR #73, opened 2026-05-24
fix: clean stop on live radio stream stall via local proxy (v2.5.1)

hoorna:
## Problem

When playing internet radio via Roon (which proxies the station through a local HTTP URL), the renderer hangs permanently after a buffer underrun if the upstream station stops sending audio.

Roon's proxy keeps the TCP connection alive with keepalives, so `av_read_frame()` blocks indefinitely — the existing FFmpeg I/O timeout does not fire because there is no socket-level timeout. The ring buffer stays empty, `getNewStream()` stays in rebuffering mode, the Diretta Link LEDs keep blinking, and the Roon playing-time counter keeps running. No UPnP command has any effect.

## Root cause

`av_read_frame()` has no way to be interrupted from outside once it is blocking. FFmpeg's `AVIOInterruptCB` mechanism exists precisely for this: it is polled periodically during blocking I/O and aborts with `AVERROR_EXIT` when it returns 1.

## Fix

- Register an `AVIOInterruptCB` on the `AVFormatContext` before `avformat_open_input()`
- In each `readSamples()` call, write a deadline (`now + 20s`) to an atomic before `av_read_frame()` and clear it after — zero-cost during normal playback (deadline == 0, first branch returns immediately)
- On `AVERROR_EXIT`, set `m_readTimeout = true` (mirrors v2.4.5 `m_decodeError`)
- In `process()`, detect `hasReadTimeout()` before the `samplesRead == 0` guard and trigger an immediate clean stop via the shared `triggerFatalStop` lambda (also used for `hasDecodeError()`)
- Teardown order: preload thread joined → next-track state cleared → `state = STOPPED` → `m_trackEndCallback()` — identical to v2.4.5

## Design

Follows the v2.4.5 corrupt-packet fix (PR #72) exactly: separate flag, reset in `open()`, accessor method, immediate teardown in `process()`, state set before callback, no drain delay. The `triggerFatalStop` lambda eliminates the duplicate 14-line teardown that would otherwise exist for both fatal conditions.

## Test

Reproduced with Mother Earth Radio Classic via Roon on a buffer underrun (~30 min playback). After the fix:
- At t+18s after underrun: `[AudioDecoder] Read timeout — stream stalled (live proxy?)`
- Immediate clean stop: Diretta target released, Roon playing-time counter freezes, stop button visible
- No hang, no blinking LEDs

Observable result is identical to the v2.4.5 corrupt-packet fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [2.5.1] - 2026-05-24
+
+### Fixed
+- **Permanent hang on live radio stream stall via Roon proxy** (reported by hoorna/Alfred for Mother Earth Radio Classic via Roon): When Roon serves an internet radio station via its local HTTP proxy, the TCP connection stays alive (keepalives) even when the upstream CDN or station stops sending audio. `av_read_frame()` blocks indefinitely on such a connection — the existing 30-second FFmpeg I/O timeout does not fire because TCP keepalives prevent a socket-level timeout. The ring buffer empties, `getNewStream()` enters rebuffering mode, and the renderer never recovers: Diretta Link LEDs keep blinking, the Roon playing-time counter runs on, and no UPnP command has any effect. The fix registers an `AVIOInterruptCB` on the `AVFormatContext` before `avformat_open_input()`. In each `readSamples()` call, a per-call deadline (`now + 20 s`) is written to an atomic before `av_read_frame()` and cleared to zero after it returns. The callback checks the deadline on every FFmpeg I/O poll; when `now >= deadline` it returns 1, causing `av_read_frame()` to abort with `AVERROR_EXIT`. A new `m_readTimeout` flag (analogous to v2.4.5's `m_decodeError`) is set, and `process()` detects it before the `samplesRead == 0` guard and triggers an immediate clean stop: preload thread joined, next-track state cleared, `m_state = STOPPED`, `m_trackEndCallback()` fired. The UPnP controller (Roon) receives the correct STOPPED state, the playing-time counter freezes, and the Diretta target is fully released — identical observable behaviour to the v2.4.5 corrupt-packet fix. The interrupt callback is zero-cost during normal playback (deadline is 0, first branch returns immediately). The `triggerFatalStop` lambda in `process()` is shared between the decode-error and read-timeout cases, eliminating the duplicate 14-line teardown.
+
 ## [2.5.0] - 2026-05-23
 
 ### Added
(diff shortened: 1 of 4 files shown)


hoorna:
Note on the failing checks: these failures are pre-existing and unrelated to this PR's changes.

- **Build (x64 / arm64)**: fail because the workflow uses `actions/upload-artifact: v3`, which GitHub has deprecated and now auto-fails. Not caused by our code.
- **Code Quality Check**: fails on trailing whitespace in `src/DirettaOutput.cpp` — a file not touched by this PR.

Both checks would fail on any PR against the current `main`.

cometdom:
Hi @hoorna, thanks for another excellent PR — really clean work. 🙏

Squash-merged into `main` (commit `71bd239`) and **published as v2.5.1** just now: https://github.com/cometdom/DirettaRendererUPnP/releases/tag/v2.5.1.

A few things I particularly enjoyed reviewing:

- **`AVIOInterruptCB` is exactly the right tool** for this — there is genuinely no other way to interrupt a blocking `av_read_frame()` from outside. The atomic deadline pattern (set before, clear after) is elegant and is *zero-cost during normal playback* — the callback hits the `if (deadline == 0) return 0;` branch and returns instantly.
- **The `triggerFatalStop` lambda refactor is a really nice touch.** Factoring the 14-line teardown between `hasDecodeError()` and `hasReadTimeout()` means future fatal-stop conditions (if any) can be added in one line each, and the teardown ordering can never drift between them. That's the kind of consolidation I would have asked for myself.
- The full **symmetry with the v2.4.5 corrupt-packet fix** (separate flag, reset in `open()`, accessor, check before `samplesRead == 0`, state-then-callback, no drain delay, both DSD-raw and PCM paths covered) made the review fast and confident.

Two **purely cosmetic** doc completions I applied post-merge (commit `493edfb`, no code change):

1. README updated to v2.5.1 (title, version badge, *What's New* rewritten around the stall fix, *Last updated* date). I also slipped the missing v2.5.0 entry into the *Previous Versions* table — it was the current *What's New* when v2.5.0 shipped so it had never been moved down.
2. Added a line to the `docs/CLAUDE.md` completed list, in the same format as the v2.4.5 / v2.5.0 entries (rationale, mechanism, scope, thread-safety note).

So this time the merge needed no actual corrections to your work — you nailed it.

Two non-blocking thoughts for possible future work, purely for discussion — not asks:

- The interrupt callback is also active during `avformat_open_input()` (since it's registered before the open). It's inoffensive today because `m_readDeadlineNs == 0` then, but the same mechanism could one day cover a stalled HTTP connect by setting a deadline around the open call too.
- `READ_STALL_TIMEOUT_S = 20` could become configurable (`--read-timeout-s` / config file) if some users with very slow upstream connections turn out to want a different value. Probably not needed in practice — 20 s is a sensible default for radio.

Thanks again — and a small wink to our shared collaborator. 🤖 Looks like our two Claudes work well together too.

PR #75, opened 2026-06-16
fix(webui): dedup config saves, preserve commented example lines

hoorna:
## Summary

`ShellVarConfig.save()` in the webui treated commented-out example
lines (e.g. `#CPU_AUDIO=2`) as if they were active settings, and
rewrote every occurrence of a key per save — so
`/etc/default/diretta-renderer` grew duplicate `KEY=VALUE` entries on
every webui save.

Two related changes plus a unit test:

- **`webui/config_parser.py`** — only match active (uncommented)
  assignment lines, update only the first occurrence per key, drop
  later active duplicates.
- **`install.sh`** — limit the migration sed to the first match per
  key with `0,/pattern/{s|...|}`. The old form activated every
  `#KEY=...` example line in the new config, which is the input the
  broken `save()` then mishandled.
- **`webui/test_config_parser.py`** — regression coverage (stdlib
  `unittest`, no new deps).

## How this was discovered

Investigating why `--cpu-other 0` thread-pinning failed on a
Raspberry Pi 4 setup. The pinning failure itself had a separate root
cause in `fedora-rpi-audiophile-setup` (follow-up PR planned there),
but `/etc/default/diretta-renderer` showing three identical
`CPU_AUDIO=2` lines after a single webui save turned out to be an
independent bug worth its own fix.

## Test plan

- [x] **Unit test** — 5/5 pass: commented-line preservation,
  duplicate collapse, idempotency, append-new-key, comment/blank
  preservation.
- [x] **Webui save cycle** on Fedora 44 / Raspberry Pi 4 — saves
  produce exactly one active line per key; repeated saves idempotent;
  `#KEY=...` example lines preserved.
- [x] **Fresh `install.sh`** on the same Pi via
  fedora-rpi-audiophile-setup — `/etc/default/diretta-renderer` ends
  up with 33 active keys, highest occurrence count is 1.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1557,9 +1557,14 @@ WRAPPER_EOF
 
                 # Check if this key is known in current version
                 if echo "$KNOWN_KEYS" | grep -qw "$key"; then
-                    # Apply to new config: replace commented or uncommented line
+                    # Apply to new config: replace only the FIRST occurrence of
+                    # the key (commented or uncommented). The 0,/pattern/ range
+                    # limits the substitution to the first match — without it,
+                    # sed would activate every '#KEY=...' example line, leaving
+                    # duplicate active entries that the webui save logic then
+                    # had to deduplicate on every subsequent save.
                     if sudo grep -q "^#\?${key}=" "$CONFIG_FILE" 2>/dev/null; then
-                        sudo sed -i "s|^#\?${key}=.*|${key}=${val}|" "$CONFIG_FILE"
+                        sudo sed -i "0,/^#\?${key}=/{s|^#\?${key}=.*|${key}=${val}|}" "$CONFIG_FILE"
                         migrated_keys="$migrated_keys $key"
                     fi
                 else
--- a/webui/config_parser.py
+++ b/webui/config_parser.py
@@ -46,9 +46,18 @@ class ShellVarConfig:
     def save(path, settings):
         """Rewrite config file, updating existing keys and preserving structure.
 
-        - Lines with existing keys are updated in place
-        - Comments and blank lines are preserved
-        - Keys not present in the file are appended at the end
+        - Active lines with known keys are updated in place (first occurrence only)
+        - Comments and blank lines are preserved unchanged
+        - Duplicate active key lines after the first are dropped (deduplication)
+        - Keys not present as active lines are appended at the end
+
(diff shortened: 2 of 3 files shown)


hoorna:
Heads-up on the CI failures — none of them are caused by this PR's changes. A quick triage:

**`Build on x64` / `Build on arm64`** — auto-failed by GitHub because the workflow uses `actions/upload-artifact@v3`, which was deprecated on 2026-04-16. The workflow fails before reaching any compile step (~3-4 s). Fix is a one-line bump to `@v4` in the workflow file.

**`Code Quality Check`** — the `Check file formatting` step greps for trailing whitespace across all `*.cpp`/`*.hpp` files. The 100+ hits all come from `src/DirettaOutput.cpp`, which this PR does not touch. Verified upstream `main` already has 286 lines with trailing whitespace in that file (`git grep -c '[[:blank:]]$' main -- src/DirettaOutput.cpp` → 286).

The three files this PR changes (`install.sh`, `webui/config_parser.py`, `webui/test_config_parser.py`) are all clean.

Happy to leave this to you to address as you see fit (separate maintenance PRs, etc.) — let me know if you'd prefer me to do anything from this side.

hoorna:
Re-tested both halves on a Fedora 44 / Raspberry Pi 4 against the latest branch tip.

**Quick gate (unit tests):**
```
$ python3 webui/test_config_parser.py
... 5 tests passed ...
```

**Real-conditions test (webui save dedup):**
- Installed `webui/config_parser.py` over the running webui at `/opt/diretta-renderer-upnp/webui/`, `systemctl restart diretta-renderer-webui`.
- Toggled `CPU_AUDIO` between 2 and 3 four times via the web UI, hitting **Save & Restart** each time.
- After those eight saves, `grep -E "^[A-Z_]+=" /etc/default/diretta-renderer | cut -d= -f1 | sort | uniq -c | sort -rn` shows every key with count 1. No accumulation. ✓
- Existing header comments preserved.

**`install.sh` sed fix:**
Validated earlier this week on a fresh install via `setup.sh` option 12 — the migrated config had 33 active keys with no duplicates and no accidentally-activated `#KEY=...` example lines. ✓

Side note: on systems where the unfixed `install.sh` had already activated `#KEY=...` example lines, the new `save()` will dedup them away — at that point they're indistinguishable from real settings. Cosmetic only, and fresh installs after this merges keep examples intact.

LGTM from this side — happy whenever it suits you.

PR #77, opened 2026-06-17
fix(tuner): reconcile slice cpuset to --cpu-* at runtime (keeps strict isolation)

cometdom:
## Summary

Alternative to #76, addressing the **same root cause** with a tighter design.

The CPU tuner generates a systemd slice with `AllowedCPUs=${RENDERER_CPUS}` (the isolated renderer cores only). DRUP then pins its own threads via `--cpu-audio` / `--cpu-decode` / `--cpu-other` (#68, v2.4.2). When any of those references a core **outside** the slice cpuset — typically a housekeeping core, e.g. `CPU_OTHER=0` — the kernel rejects the affinity call with **`EINVAL`**. On top of that, the `ExecStartPost=` round-robin `distribute-diretta-threads.sh` overwrote whatever pinning *did* land.

## Why a separate PR instead of #76

[#76](https://github.com/cometdom/DirettaRendererUPnP/pull/76) widens `AllowedCPUs` to **all** cores at install time (`HOUSEKEEPING_CPUS,RENDERER_CPUS`). That is fail-safe and reconfig-robust, but it **permanently relaxes the slice**: strict isolation is lost, and on x86 the **SMT siblings** of the housekeeping core are pulled into the cpuset — even for users who never set a `--cpu-*` flag.

The cgroup `AllowedCPUs` is **static** (baked at install) while the `CPU_*` settings are **dynamic** (changed anytime via the web UI, then a restart). A tuner that *read* `CPU_*` at install time would be tighter but would break on the next reconfig (EINVAL again), unless the web UI were coupled to re-run the commit-pinned tuner — an undesirable coupling.

## Approach

Move `--cpu-*` awareness to **`start-renderer.sh`** — the only component that already knows `CPU_AUDIO`/`CPU_DECODE`/`CPU_OTHER` **and** runs on every (re)start (so it always matches the live config).

- **`diretta-renderer-tuner.sh` / `-nosmt`**: slice stays **strict** (`AllowedCPUs=${RENDERER_CPUS}`, unchanged from `main`); **drop the conflicting `ExecStartPost`** — this is the correct half of #76 and is kept here.
- **`systemd/start-renderer.sh`**: just before `exec`, if any `CPU_*` is set *and* the service is confined to a cpuset slice, widen that slice at runtime via `systemctl set-property --runtime` to exactly **`renderer cores + referenced CPU_* cores`**.

## Result vs #76

| Scenario | #76 | This PR |
|---|---|---|
| No `--cpu-*` flags | slice = all cores (isolation lost; SMT siblings leak on x86) | **strict isolation preserved** |
| With `--cpu-*` flags | works, but cpuset includes extra cores | works; cpuset = **exact intent** |
| Reconfig via web UI | robust (static superset) | robust (**recomputed every start**) |

## Trade-off (honest)

Unlike #76 (fail-safe by construction), the `CPU_OTHER=0` case now depends on `set-property` succeeding. On failure a clear `WARNING` is logged and behaviour degrades to the pre-#76 EINVAL **for that specific pinning only**. `set-property AllowedCPUs` has been stable on the targeted Fedora/systemd versions for years, so the practical risk on the wizard's targets is negligible.

## Relation to other PRs

- Supersedes #76 (keeps its `ExecStartPost` removal; replaces its slice change). Suggest closing #76 in favour of this once validated.
- Independent of #75.
- Note for @hoorna: same root-cause analysis as your #76 — this just relocates the fix so strict isolation survives when no `--cpu-*` flags are used. Happy to discuss.

## Test plan

- [x] `bash -n` on `diretta-renderer-tuner.sh`, `diretta-renderer-tuner-nosmt.sh`, `systemd/start-renderer.sh`.
- [x] Functional simulation of `reconcile_cpu_slice` (fake `systemctl`): Pi-4 full-pin (`CPU_AUDIO=2 CPU_DECODE=3 CPU_OTHER=0` → cpuset `{0,1,2,3}`), x86 partial-pin (`CPU_AUDIO=3,4` → `{1,2,3,4}`, no SMT-sibling leak), no-flags (slice untouched).
- [ ] **Not yet on hardware** — needs a Pi 4 and an x86 SMT box. Bump the wizard's pinned tuner commit (`fedora-*-audiophile-setup` module 02) only after that.


cometdom:
@hoorna — if you're up for testing this on your Pi 4 before we merge, here's a focused test memo. Two nuances first, then the commands.

### Nuance 1 — `git checkout` does not deploy
Everything runs from `/opt/diretta-renderer-upnp/` (installed by `install.sh`), not from the checkout, and the tuner is fetched by the wizard from a pinned commit. So after checking out a branch you must re-inject the changed files into `/opt` (or re-run `install.sh`), otherwise the running service is still on the old code.

### Nuance 2 — on a Pi 4, #77 is equivalent to #76
The Pi 4 has no SMT, so #77's reconciled cpuset becomes `0-3` = all 4 cores = exactly what #76 produces. So on your hardware you can confirm **"EINVAL gone, pinning works, no regression"**, but the actual *benefit* of #77 over #76 (strict isolation preserved when no `--cpu-*` flag is set, and no SMT-sibling leak) is only observable on an x86 box with SMT — Dominique will cover that side.

---

### Quick gate (anywhere, no deploy) — the #75 unit tests
```bash
git checkout fix/webui-config-save-dedup
python3 webui/test_config_parser.py     # expect 5/5 OK
```

### #75 (webui dedup) in real conditions
```bash
git checkout fix/webui-config-save-dedup
sudo cp webui/config_parser.py /opt/diretta-renderer-upnp/webui/
sudo systemctl restart diretta-renderer-webui
# In the web UI: save a CPU_* setting 2-3 times, then:
cat /etc/default/diretta-renderer       # exactly ONE active line per key; '#...' examples intact
```

### #77 (tuner + runtime cpuset reconciliation)
```bash
git checkout feat/tuner-aware-cpu-flags
sudo ./diretta-renderer-tuner.sh apply                 # writes strict slice + drop-in (no ExecStartPost)
sudo cp systemd/start-renderer.sh /opt/diretta-renderer-upnp/start-renderer.sh
# /etc/default/diretta-renderer:  CPU_AUDIO=2  CPU_DECODE=3  CPU_OTHER=0
sudo systemctl daemon-reload && sudo systemctl restart diretta-renderer
journalctl -u diretta-renderer -b | grep -iE "Reconciling|EINVAL|affinity"
ps -p "$(systemctl show -p MainPID --value diretta-renderer)" -L -o tid,psr,comm
```
What to look for on the Pi 4:
- a `Reconciling diretta-renderer.slice AllowedCPUs: renderer cores [...] + CPU_* [...]` line in the journal,
- **no `EINVAL`**, audio thread on CPU 2, decode on CPU 3, other threads incl. CPU 0,
- no competing redistribution (the `ExecStartPost` is gone).

When you're done, `git checkout main` and re-run `install.sh` (or copy the original files back) to return to the released code.

No rush at all — this is exactly the "let's not move too fast" pass before we merge. Thanks for the original root-cause work on #75/#76; the runtime-reconciliation idea here is built directly on your analysis.

hoorna:
Tested this PR on a Fedora 44 / Raspberry Pi 4. Started from the upstream-baked state (the wizard had just reinstalled DRUP, so slice `AllowedCPUs=1-3` and the ExecStartPost line were both back). Replaced `start-renderer.sh` with this PR's version, and removed the `ExecStartPost=` line from the drop-in by hand to mirror this PR's tuner edits.

## What works

**With `CPU_AUDIO=2 CPU_DECODE=3 CPU_OTHER=0`:**

- Reconcile logs as advertised: `Reconciling diretta-renderer.slice AllowedCPUs: renderer cores [1-3] + CPU_* [2 3 0]`
- `systemctl show -p AllowedCPUs --value diretta-renderer.slice` → `0-3`
- 8 "other" threads on CPU 0, SDK worker on CPU 2, decode on CPU 3, nothing on CPU 1.

Same end result as #76, but the cpuset is only widened to what the user actually asked for. ✓

**With all `CPU_*` empty, after a fresh reboot:**

- No reconcile log line (function bails early, as designed).
- `AllowedCPUs` stays at `1-3`.
- Threads schedule across cores 1, 2, 3 only — no thread on CPU 0.

Strict isolation fully preserved when no `--cpu-*` flags are set. ✓

## One edge case I'd want to flag

The runtime override that `set-property --runtime` installs in the first scenario **persists across `systemctl restart diretta-renderer`** — it's only cleared by a full systemd restart or a reboot. So if a user transitions from "with `CPU_*`" to "without `CPU_*`" *within the same boot* (e.g. clears them in the web UI and clicks restart), strict isolation isn't restored: the slice keeps the widened cpuset from the previous run.

Reproduced:
1. Boot with `CPU_*` set → slice becomes `0-3`, threads land correctly.
2. Comment out all `CPU_*`, `systemctl restart diretta-renderer`.
3. `reconcile_cpu_slice` runs, sees no `CPU_*`, returns at the early guard without touching the slice.
4. `AllowedCPUs` is still `0-3` from step 1. Threads cluster on CPU 0.
5. Only `sudo reboot` (or an explicit `set-property --runtime diretta-renderer.slice AllowedCPUs=1-3`) restores strict isolation.

The comment on the early-return reads *"leave the strict tuner cpuset as-is"*, but if a previous start already widened it at runtime, "as-is" no longer means "strict". The function reads the effective value via `systemctl show`, which reflects the runtime override and not the file-baked value.

In normal day-to-day use this is probably rare (users don't often remove `CPU_*` mid-boot), but for someone who's debugging or reconfiguring via the web UI it would be surprising.

## Two possible directions (suggesting, not prescribing)

1. **On the no-`CPU_*` path, explicitly reset to the tuner-baked value.** Read it from `/etc/systemd/system/diretta-renderer.slice` (or `systemctl cat`), then `set-property --runtime <slice> AllowedCPUs=<that>`. One extra call per start in the no-flags case.
2. **Cache the original baked value on the first reconcile** and use that as the "reset target" on subsequent runs.

Either way the goal would be: every service start brings the slice cpuset to a known state regardless of what the previous start did.

## On keeping #76 open as a fallback

Agree with that plan. If the edge case above is fixable in #77 without much extra surface area, this PR supersedes #76 cleanly. If you'd rather ship the fail-safe path first and revisit runtime reconciliation later, #76 is also fine on Pi 4. Happy to follow your judgement on which way to go.

Thanks for the rework — the diagnosis ("static install-time vs. dynamic runtime config") is the right one. Pinning that into `start-renderer.sh` rather than the tuner is the cleaner place.

cometdom:
Great catch @hoorna — that's a real bug, not a corner I want to hand-wave. You nailed both the mechanism (`set-property --runtime` survives `systemctl restart`, only a reboot/daemon-reexec clears it) and why the early-return was wrong: it read the *effective* `AllowedCPUs`, which already carried the previous start's widening, so "as-is" silently meant "still widened".

Fixed in `50538b9` along the lines of your option 1, with one tweak — instead of hardcoding the slice path I read the baked value from the unit file via `FragmentPath`, which is immune to `--runtime` drop-ins:

- base = `sed -n 's/^AllowedCPUs=//p' "$(systemctl show -p FragmentPath --value <slice>)" | tail -n1`
- every start now **re-asserts a deterministic target**: `base` alone when no `CPU_*` is set (restores strict isolation), `base + CPU_*` otherwise.

So the slice reaches a known state on every start regardless of the previous run. Cost is one extra `set-property` per start in the no-flags case, exactly as you anticipated — cheap and worth it.

Simulated the transition you reproduced:
```
start 1  CPU_AUDIO=2 CPU_DECODE=3 CPU_OTHER=0  -> set AllowedCPUs = 1-3 2 3 0   (=> 0-3)
start 2  (same boot) all CPU_* cleared          -> set AllowedCPUs = 1-3         (strict restored)
x86      CPU_AUDIO=3,4 only                      -> set AllowedCPUs = 1-3 3 4     (no SMT-sibling leak)
```

Would you mind re-running your exact repro on the Pi 4 against `50538b9`? Specifically step 2→4: clear `CPU_*`, `systemctl restart`, then confirm `systemctl show -p AllowedCPUs --value diretta-renderer.slice` is back to `1-3` and no thread sits on CPU 0. If that holds on real hardware, this supersedes #76 cleanly and we can close it.

Thanks again — the "every start brings the slice to a known state" framing was the right spec.

hoorna:
Confirmed on Fedora 44 / Raspberry Pi 4 against `50538b9`. The repro scenario now passes — log lines speak for themselves:

```
Jun 17 22:28:06 Fedora diretta-renderer[1705]: Reconciling diretta-renderer.slice AllowedCPUs: renderer cores [1-3] + CPU_* [2 3 0]
Jun 17 22:28:10 Fedora diretta-renderer[1783]: Reconciling diretta-renderer.slice AllowedCPUs to baked renderer cores [1-3] (no CPU_* set)
```

Step-by-step within one boot:

- Start with `CPU_AUDIO=2 CPU_DECODE=3 CPU_OTHER=0` → slice widens to `0-3` (first log line). ✓
- Comment out all `CPU_*`, `systemctl restart diretta-renderer` (no reboot) → the new "to baked renderer cores [1-3] (no CPU_* set)" log line appears, `systemctl show -p AllowedCPUs --value diretta-renderer.slice` returns **`1-3`**, threads land only on CPU 1/2/3 — none on CPU 0. ✓

Reading `base` from `FragmentPath` is exactly the right design choice — the cpuset is now genuinely independent of whatever the previous start did. Happy for #76 to be closed in favour of this.

Thanks for the quick turnaround, and for the credit comment in the code.

cometdom:
Perfect — thanks for the thorough re-test, @hoorna. The two log lines tell the whole story, and `AllowedCPUs` returning to `1-3` after the no-reboot restart is exactly the deterministic behaviour we were after.

So #77 is now validated on both ends: x86 + SMT (Dominique) and Fedora 44 / Pi 4 (you), edge case included. Plan from here, pending Dominique's call on the actual merges/releases:

- merge #75 (webui dedup) and #77 (this),
- close #76 in favour of #77,
- mirror the #75 fix in slim2Diretta (already staged),
- cut releases and bump the pinned `_TUNER_COMMIT` in both `fedora-*-audiophile-setup` wizards.

Really appreciate the careful root-cause work on #75/#76 and the precise edge-case repro here — that's what made the runtime-reconciliation approach land cleanly.

cometdom:
Shipped 🎉 — thanks @hoorna, this is all live now.

- **DirettaRendererUPnP v2.5.4** (this PR #77 + your #75): https://github.com/cometdom/DirettaRendererUPnP/releases/tag/v2.5.4
- **slim2Diretta v1.4.7** (symmetric port of #75): https://github.com/cometdom/slim2Diretta/releases/tag/v1.4.7
- Both `fedora-audiophile-setup` and `fedora-rpi-audiophile-setup` have had their pinned tuner commit bumped to `fa3ea43`, so fresh installs pick up the strict-slice + runtime-reconciliation behaviour.

Your root-cause work on #75/#76 and the precise Pi 4 edge-case repro are what made the runtime-reconciliation design land cleanly — both are credited in the v2.5.4 changelog. Really appreciate the careful testing throughout. 🎧

hoorna:
@cometdom, you're welcome!

PR #84, opened 2026-08-10
fix(stream): recover from live stream proxy stalls without stopping

hoorna:
## Summary

- Registers an `AVIOInterruptCB` that aborts `av_read_frame()` after a configurable stall timeout (`READ_STALL_TIMEOUT_S = 5s`)
- On stall, attempts up to 3 reconnects before giving up with a clean stop
- During normal playback the deadline is 0 — the callback returns in a single atomic load (nanosecond overhead, no effect on audio quality)
- Covers both PCM and DSD raw paths

## Motivation

When a stream is proxied via a local HTTP server (e.g. Roon, Audirvana), the TCP connection stays alive via keepalives even when the upstream stops delivering audio data. `av_read_frame()` then blocks indefinitely, hanging the renderer permanently with no way to recover via UPnP commands.

The timeout mechanism applies to all streams but only triggers on an actual stall. In practice a 5-second stall only occurs when the upstream source (e.g. a live radio relay) stops delivering data — for local files and Qobuz/Tidal streams the reads complete in microseconds and the timeout never fires.

## Test plan

- [x] Verified live stream reconnects cleanly after a proxy stall
- [x] Verified no regression on normal local file playback (pcm_s16le, pcm_s24le)
- [x] Verified reconnect succeeds on intermittent stalls (6 reconnects over 12-hour overnight test, all on attempt 1/3)

### DIFF
--- a/src/AudioEngine.cpp
+++ b/src/AudioEngine.cpp
@@ -2266,7 +2266,7 @@ bool AudioEngine::process(size_t samplesNeeded) {
         // On a live stream the upstream server may reset its connection, causing the
         // local Roon proxy to stall briefly before resuming. Reopen the same URL
         // instead of stopping — but cap retries to catch a truly dead stream.
-        if (m_currentTrackInfo.duration == 0 && m_liveStreamReconnects < 3) {
+        while (m_currentTrackInfo.duration == 0 && m_liveStreamReconnects < 3) {
             ++m_liveStreamReconnects;
             std::cerr << "[AudioEngine] Live stream stalled, attempting reconnect ("
                       << m_liveStreamReconnects << "/3)..." << std::endl;
@@ -2275,7 +2275,7 @@ bool AudioEngine::process(size_t samplesNeeded) {
                 std::cout << "[AudioEngine] Reconnect successful, resuming live stream" << std::endl;
                 return true;
             }
-            std::cerr << "[AudioEngine] Reconnect failed" << std::endl;
+            std::cerr << "[AudioEngine] Reconnect failed (attempt " << m_liveStreamReconnects << "/3)" << std::endl;
         }
         triggerFatalStop("Stream read timeout (live proxy stall), triggering clean stop");
         return false;
--- a/src/AudioEngine.cpp
+++ b/src/AudioEngine.cpp
@@ -1186,7 +1186,7 @@ size_t AudioDecoder::readSamples(AudioBuffer& buffer, size_t numSamples,
             // Read packets until we have enough data
             // DSF layout: each packet is [blockSize L][blockSize R]
             while (leftOffset < bytesPerChannelNeeded && !m_eof && !m_readTimeout) {
-                // Deadline: abort if av_read_frame() blocks > 20s (live stream proxy stall)
+                // Deadline: abort if av_read_frame() blocks > 5s (live stream proxy stall)
                 m_readDeadlineNs.store(
                     std::chrono::duration_cast<std::chrono::nanoseconds>(
                         (std::chrono::steady_clock::now() + std::chrono::seconds(READ_STALL_TIMEOUT_S)).time_since_epoch()
@@ -1354,7 +1354,7 @@ size_t AudioDecoder::readSamples(AudioBuffer& buffer, size_t numSamples,
     }
 
     while (totalSamplesRead < numSamples && !m_eof && !m_readTimeout) {
-        // Read packet — set 20s deadline so av_read_frame() cannot block indefinitely
+        // Read packet — set 5s deadline so av_read_frame() cannot block indefinitely
         // (protects against live streams via proxy keeping TCP alive with no audio data)
         m_readDeadlineNs.store(
             std::chrono::duration_cast<std::chrono::nanoseconds>(
@@ -1955,6 +1955,7 @@ bool AudioEngine::play() {
     // Open current track if not already open OR if at EOF
     if (!m_currentDecoder || m_currentDecoder->isEOF()) {
         std::cout << "[AudioEngine] Opening track (new or after EOF)" << std::endl;
+        m_liveStreamReconnects = 0;
 
         if (!openCurrentTrack()) {
             std::cerr << "[AudioEngine] Failed to open track" << std::endl;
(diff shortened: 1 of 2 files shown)


cometdom:
Reviewed this PR (high-effort pass, with verification). One confirmed blocking issue plus several secondary findings.

**Confirmed: mutex held across blocking reconnect (deadlock-adjacent)**
`AudioEngine::process()` takes `std::unique_lock<std::mutex> lock(m_mutex)` and never releases it before calling `openCurrentTrack()` on the reconnect path (unlike the preload path and `triggerFatalStop`, which both `unlock()` first). `openCurrentTrack()` → `AudioDecoder::open()` → `avformat_open_input()` with a 10–30s FFmpeg timeout. Worst case: ~90s (3 attempts × 30s) with `AudioEngine::m_mutex` held.

Meanwhile `onSetURI`/`onPlay`/`onStop`/`onPause`/`onSeek` in `DirettaRenderer` all take `DirettaRenderer::m_mutex` and then block (non-`try_lock`) on `AudioEngine::m_mutex` via `setCurrentURI()`/`play()`. Result: while a reconnect is in flight, every UPnP transport command freezes for up to ~90s — for what's meant to be a transparent, inaudible recovery.

Suggested fix: `lock.unlock()` around the `openCurrentTrack()` call in the reconnect branch, same pattern already used elsewhere in this function.

**Other findings:**
- `openCurrentTrack()` unconditionally fires `m_trackChangeCallback` → `notifyGaplessTransition`, which resets the UPnP position to 0:00 and re-announces the track to control points on *every* reconnect attempt (up to 3x per stall). Since the URI/metadata haven't changed, this is a visible glitch for what should be silent recovery.
- `READ_STALL_TIMEOUT_S` 20s→5s combined with the `duration == 0` heuristic can also match malformed/NOPTS local files (network mounts, USB spin-up stalls), causing a silent restart-from-zero instead of a clean stop on what used to just be a slow local read.
- `m_liveStreamReconnects` is never reset when a *new* track is opened — only on a successful read. A stream that exhausts its 3 attempts and gets replaced by a new URI that also stalls before any successful read inherits the exhausted counter and skips straight to `triggerFatalStop`.
- Two independent, inconsistent recovery mechanisms now exist for "live stream hiccup": `m_liveStreamDecodeErrors` (decoder-level, cap 1, flush+continue) and `m_liveStreamReconnects` (engine-level, cap 3, full reconnect). They don't compose — a decode-error path goes straight to `triggerFatalStop` without ever getting a reconnect attempt.
- 4 leftover "20s" comments not updated to match the new 5s timeout (`AudioEngine.h:194`, `AudioEngine.cpp:1188,1356,2251`).
- No backoff between reconnect attempts — a fast-failing connect (DNS/refused) could burn all 3 attempts near-instantly.
- Minor: `m_liveStreamDecodeErrors < 1` / `m_liveStreamReconnects < 3` are inline magic numbers rather than named `constexpr`, inconsistent with the rest of the file (`READ_STALL_TIMEOUT_S` etc.).

The core recovery idea is good and the interrupt-callback zero-cost design is solid — the mutex issue is the one thing I'd want fixed before merge.

cometdom:
Thanks for the fast turnaround — the `if`→`while` retry-count bug, the mutex deadlock, the spurious gapless notification, the `m_liveStreamReconnects` per-track reset, and the stale comments are all correctly addressed by the two new commits.

However, the mutex fix (`lock.unlock()` around `openCurrentTrack()` in the reconnect loop) introduces a new, more serious problem: a data race on `m_currentDecoder`.

`openCurrentTrack()` (`AudioEngine.cpp:2375-2394`) writes to `m_currentDecoder`, `m_currentURI`, and `m_currentTrackInfo` with **no internal locking** — its own comment says "called with m_mutex held, except from the reconnect path which unlocks first." That reconnect path is exactly `process()` at `AudioEngine.cpp:2276-2278`, where `lock.unlock()` is called right before `openCurrentTrack(true)` runs, for up to ~30s per attempt (3 attempts).

During that unlocked window, if a UPnP Stop or SetURI command arrives on the other thread:
- `AudioEngine::stop()` (`AudioEngine.cpp:2006`) uses `try_to_lock`, which now **succeeds immediately** (the mutex was just released) and calls `m_currentDecoder.reset()` (`AudioEngine.cpp:2011`)
- `setCurrentURI()` (called from `onSetURI`/`onStop`, `AudioEngine.cpp:1872` `lock_guard`, blocking) also calls `m_currentDecoder.reset()` (`AudioEngine.cpp:1888`)

...while `openCurrentTrack()` is concurrently doing `m_currentDecoder = std::make_unique<AudioDecoder>()` (`AudioEngine.cpp:2386`) and calling `->open(m_currentURI)` (`AudioEngine.cpp:2388`) on the audio thread, with no synchronization at all between the two sides. This is a genuine data race on a non-atomic `unique_ptr` — undefined behavior at best, and if `stop()`'s `reset()` destroys the `AudioDecoder` object while `open()` is still executing on it from the other thread, a real use-after-free/crash.

For comparison, `preloadNextTrack()` (`AudioEngine.cpp:2414`) handles the equivalent "slow I/O without holding the lock" problem correctly with a capture-validate-commit pattern: it builds the new decoder into a **local** `unique_ptr` and only touches the shared `m_nextDecoder` member briefly under lock at commit time, after re-validating nothing changed while it was unlocked. The reconnect path in `openCurrentTrack()` doesn't follow that pattern — it mutates the shared members directly and unguarded.

Net effect: this trades a ~90s UI freeze for a potential crash if the user hits Stop (or changes track) during a live-stream stall reconnect — a quite plausible real-world sequence. Suggest reworking the reconnect path to follow the same capture-validate-commit shape as `preloadNextTrack()` before merge.

cometdom:
Verified — the capture-validate-commit fix is correct. Walked through the concurrency scenarios by hand:
- `capturedURI` is snapshotted under lock before unlocking, avoiding a torn read on the `std::string`.
- The new decoder is built entirely in a local `unique_ptr` while unlocked — no shared member touched during the slow I/O.
- On re-acquiring the lock, `m_state != State::PLAYING || m_currentURI != capturedURI` catches both a concurrent `stop()` (state is set atomically before `stop()` even attempts the mutex, so this check works regardless of whether `stop()`'s `try_to_lock` succeeded) and a concurrent `setCurrentURI()`.
- Commit to `m_currentDecoder`/`m_currentTrackInfo` only happens under lock, after validation — matches the `preloadNextTrack()` pattern exactly, no more race.

One leftover, non-blocking: `openCurrentTrack(bool suppressCallback = false)` — the reconnect path no longer calls `openCurrentTrack()` at all (logic was inlined instead), so `suppressCallback` is now dead — nothing ever passes `true`. Harmless, just an unused parameter; feel free to leave it or drop it in a follow-up.

Merged directly — GitHub's mergeability check flagged a conflict against `main` because PR #86 (raw packet bypass) merged in the meantime and both PRs add an init line at the same spot in `AudioDecoder::open()` (`m_rawPacketBypass = false;` vs `m_liveStreamDecodeErrors = 0;`). Resolved by keeping both lines — trivial, no logic overlap. Rebuilt and linked clean before pushing.

PR #85, opened 2026-08-10
fix: prevent buffer oscillation after live stream reconnect

hoorna:
## Summary

- After a reconnect following a stream stall, raises the rebuffering threshold to 50% for one cycle (instead of the normal 20%), giving the buffer a larger cushion before resuming playback
- Reverts to the 20% threshold after that first post-reconnect cycle
- Introduces `REBUFFER_THRESHOLD_REMOTE_PCT = 0.50f` constant alongside existing `REBUFFER_THRESHOLD_PCT = 0.20f`

## Motivation

After a reconnect, the stream proxy needs a moment to stabilise data delivery. Resuming at the normal 20% threshold (~52 KB) means playback restarts before the stream is stable, causing a second immediate underrun and audible gap. Waiting until 50% (~131 KB) provides enough buffer to absorb the initial delivery variance.

This fix is strictly conditional: it only activates after `consumeReconnectFlag()` returns true, which requires a successful reconnect after a 5-second stall (see PR #84). For local files and Qobuz/Tidal streams where no stall occurs, this code path is never reached.

## Dependency

Depends on PR #84 (`fix/live-stream-decode-recovery`): requires the reconnect signal (`consumeReconnectFlag()`) introduced there. Please merge PR #84 first.

## Test plan

- [x] Verified `[post-reconnect]` tag appears in log after each reconnect
- [x] Verified rebuffering threshold = 131072 on first post-reconnect cycle
- [x] Verified threshold reverts to 52428 on subsequent underruns
- [x] 12-hour overnight test: 6 reconnects, all handled correctly, no audible noise

### DIFF
--- a/src/AudioEngine.cpp
+++ b/src/AudioEngine.cpp
@@ -2266,7 +2266,7 @@ bool AudioEngine::process(size_t samplesNeeded) {
         // On a live stream the upstream server may reset its connection, causing the
         // local Roon proxy to stall briefly before resuming. Reopen the same URL
         // instead of stopping — but cap retries to catch a truly dead stream.
-        if (m_currentTrackInfo.duration == 0 && m_liveStreamReconnects < 3) {
+        while (m_currentTrackInfo.duration == 0 && m_liveStreamReconnects < 3) {
             ++m_liveStreamReconnects;
             std::cerr << "[AudioEngine] Live stream stalled, attempting reconnect ("
                       << m_liveStreamReconnects << "/3)..." << std::endl;
@@ -2275,7 +2275,7 @@ bool AudioEngine::process(size_t samplesNeeded) {
                 std::cout << "[AudioEngine] Reconnect successful, resuming live stream" << std::endl;
                 return true;
             }
-            std::cerr << "[AudioEngine] Reconnect failed" << std::endl;
+            std::cerr << "[AudioEngine] Reconnect failed (attempt " << m_liveStreamReconnects << "/3)" << std::endl;
         }
         triggerFatalStop("Stream read timeout (live proxy stall), triggering clean stop");
         return false;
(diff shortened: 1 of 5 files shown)


cometdom:
Reviewed this PR (high-effort pass). This builds on #84, so please see that review first — the mutex/deadlock issue there affects the reconnect path this PR's threshold logic sits on top of.

**Findings specific to this PR:**
- `m_postReconnectRebuffering` is only cleared where it's consumed inside `getNewStream()`. It's never reset on the same-format quick-resume path, `close()`, or `fullReset()`. If a track ends or transitions to a same-format track before the ring buffer refills to 50%, the flag survives and gets applied to the *next*, unrelated underrun — even on a stream that never reconnected — forcing the wrong 50% threshold instead of the normal 20%.
- The reconnect path doesn't reset `m_samplesPlayed`, unlike the existing decoder-null auto-reopen path a few lines above it. After a reconnect the new decoder starts reading from position 0, but `m_samplesPlayed` keeps accumulating from before the stall, so `GetPositionInfo` drifts out of sync with actual playback after every reconnect.
- The second "catch-up" call to `process()` in `DirettaRenderer.cpp` (used when the buffer is already low) doesn't check `consumeReconnectFlag()`, unlike the first call earlier in the same function. This is precisely the more likely path to hit a reconnect (buffer already depleted), so activation of the 50% threshold gets delayed by one loop iteration in the case that matters most.
- Minor: one new log line exceeds the project's 120-char convention (CLAUDE.md).

Suggest fixing the flag lifecycle (clear it wherever `m_rebuffering` itself gets reset, not just on the happy path) and syncing the two `process()` call sites before merge.

cometdom:
Verified — all four lifecycle fixes are correctly in place: `m_postReconnectRebuffering` is now cleared in `open()`, `close()`, `fullReset()`, and at the threshold-met point in `getNewStream()`, and the second "catch-up" `process()` call in `DirettaRenderer::audioThreadFunc()` now checks `consumeReconnectFlag()` too. The `m_reconnectOccurred` atomic + `consumeReconnectFlag()` exchange is clean, no leftover races.

Minor, non-blocking: the "Transient frame error..." log line (inherited from #84's changes) is still 131 chars, over the project's 120-char convention — cosmetic only.

Merged directly, same as #84 — this PR already contained #84's commits as ancestors, so once #84 landed on `main` the merge was actually conflict-free. Rebuilt and linked clean before pushing.

PR #88, opened 2026-08-23
fix(main): prevent crash from a second SIGINT/SIGTERM during shutdown

hoorna:
## Summary
- `signalHandler()` does blocking work (renderer stop, thread joins, SDK release) directly inside the signal handler, but only the thread currently running the handler has re-delivery of the same signal blocked — every other thread (UPnP, audio, position, Diretta SDK worker, etc.) stays eligible to receive a second SIGINT/SIGTERM while the first is still being handled.
- If the kernel delivers the second signal to one of those threads instead of the main thread, `signalHandler()` re-enters concurrently and calls `DirettaRenderer::stop()` a second time while the first call is still in progress, joining the same `std::thread` objects from two threads at once.
- Observed as `terminate called without an active exception` / abort after pressing Ctrl-C twice in quick succession during a shutdown that took a few seconds (SDK release, buffer drain, several thread joins).
- Fix: block SIGINT/SIGTERM on the main thread right after registering the handlers, before any worker thread is created — new threads inherit the creating thread's signal mask, so every worker thread is now born with these signals blocked and can never receive them. Unblocked again on the main thread only, right before the final wait loop, once every worker thread already exists.
- Minimal, isolated change: only `src/main.cpp`, +25 lines, no change to the handler itself or to shutdown behavior on a single signal.

## Test plan
- [x] Compiles clean (`clang++ -Wall -Wextra -fsyntax-only`, no new warnings)
- [ ] Manual test: start the renderer, press Ctrl-C twice in quick succession during shutdown, confirm no crash (previously reproducible)
- [ ] Manual test: normal single Ctrl-C / SIGTERM still shuts down cleanly as before

### DIFF
--- a/src/DirettaRenderer.cpp
+++ b/src/DirettaRenderer.cpp
@@ -23,10 +23,23 @@
 #include <pthread.h>
 #include <sched.h>
 #include <vector>
+#include <csignal>
 
 // Logging: uses centralized LogLevel system from LogLevel.h (included via DirettaSync.h)
 // DEBUG_LOG kept as alias for backward compatibility within this file
 #define DEBUG_LOG(x) LOG_DEBUG(x)
+
+// Blocks SIGINT/SIGTERM on the calling thread only. Called at the top of
+// every worker thread's entry function so the main thread stays the sole,
+// always-interruptible receiver of a process-directed signal — see the
+// matching helper and comment in main.cpp (PR #88 review, cometdom).
+static void blockShutdownSignalsOnThisThread() {
+    sigset_t blockedSignals;
+    sigemptyset(&blockedSignals);
+    sigaddset(&blockedSignals, SIGINT);
+    sigaddset(&blockedSignals, SIGTERM);
+    pthread_sigmask(SIG_BLOCK, &blockedSignals, nullptr);
+}
 
 // Parse comma-separated core list (e.g. "6,7,8") into a vector of ints.
 // Returns empty vector on parse error or empty input.
@@ -862,6 +875,7 @@ void DirettaRenderer::stop() {
 //=============================================================================
 
 void DirettaRenderer::upnpThreadFunc() {
+    blockShutdownSignalsOnThisThread();
     auto cores = parseCoreList(m_config.cpuOther);
     if (!cores.empty()) pinThreadToCores(cores, "UPnP Thread");
     DEBUG_LOG("[UPnP Thread] Started");
@@ -874,6 +888,7 @@ void DirettaRenderer::upnpThreadFunc() {
 }
 
 void DirettaRenderer::audioThreadFunc() {
+    blockShutdownSignalsOnThisThread();
     // Prefer --cpu-decode for the audio thread when set; otherwise fall back
     // to --cpu-other (legacy behaviour). When --cpu-decode is used, also raise
     // the audio thread to SCHED_FIFO so it benefits from the same real-time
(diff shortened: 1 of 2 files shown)


cometdom:
Thanks for tracking this down — the double-signal-during-shutdown race is real and the diagnosis is spot on. Reviewed the fix (high-effort pass, verified against the actual thread-creation call sites) and it compiles clean, but I think it introduces a regression that's worth fixing before merge.

**The issue:** `pthread_sigmask(SIG_BLOCK, ...)` is set on the main thread very early in `main()` (right after registering the handlers, before `g_renderer->start()` is ever called), and `SIG_UNBLOCK` only happens *after* `start()` returns successfully. But `DirettaRenderer::start()` contains its own indefinite retry loops — waiting for the network to be ready, and waiting for the Diretta target to be discovered (both documented, intentional resilience features, not edge cases). Those loops are cancelled *only* via `stopSignal`/`g_running`, which is *only* ever set by `signalHandler()`.

During that entire retry window, every thread that exists (main, plus `g_logDrainThread` in verbose mode) has SIGINT/SIGTERM blocked — none of them have it unblocked. Since a process-directed signal is only delivered to a thread that doesn't have it blocked, and no such thread exists during this window, the signal just sits pending at the process level. Concretely: if a user hits Ctrl-C (or systemd sends SIGTERM via `systemctl stop`) while DRUP is waiting for its Diretta target to power on or the network to come up, nothing happens — `signalHandler()` never runs — until the retry eventually succeeds on its own, or the process is SIGKILL'd. Previously (pre-PR) this window was interruptible since nothing was blocking the signal on main.

**Suggested fix:** don't block SIGINT/SIGTERM on the main thread's own mask across the whole startup sequence — main should stay interruptible for its entire lifetime, that's the one thread we always want handling the signal. The actual requirement is only that *worker* threads never independently receive it, which can be achieved either by:
- bracketing the block/unblock tightly around each individual `std::thread(...)` call instead of the whole `start()` duration (4 call sites: `g_logDrainThread` in `main.cpp`, plus `m_upnpThread`/`m_audioThread`/`m_positionThread` in `DirettaRenderer.cpp`), or
- blocking SIGINT/SIGTERM at the top of each worker thread's own entry function instead of relying on inherited mask from a temporarily-blocked main.

Either keeps main permanently interruptible (fixing the regression) while still guaranteeing worker threads can never re-enter the handler (keeping your original fix intact).

Happy to take a pass at this myself if useful, or let me know if I'm missing something about how the retry loop is meant to be cancelled.

hoorna:
Thanks for the thorough review — that regression is exactly right, and I hadn't considered `start()`'s own indefinite retry loops when picking the block/unblock window.

Pushed a fix using your second suggested approach: the main thread never blocks SIGINT/SIGTERM now (so it stays interruptible for its entire lifetime, including during those retry loops). Each worker thread instead blocks the signals on itself as the first thing in its own entry function — `g_logDrainThread` in `main.cpp`, and `upnpThreadFunc`/`audioThreadFunc`/`positionThreadFunc` in `DirettaRenderer.cpp` (the same 4 call sites you identified). This still guarantees a second signal mid-shutdown can never land on a worker thread and re-enter `signalHandler()` concurrently (the original crash), without the un-interruptible window.

cometdom:
Confirmed — this correctly addresses the regression. Verified both statically and by actually running it:

- Main thread never blocks SIGINT/SIGTERM now, so it stays interruptible for the process's entire lifetime, including during `start()`'s indefinite retry loops.
- Each worker thread (`upnpThreadFunc`, `audioThreadFunc`, `positionThreadFunc`, `g_logDrainThread`) blocks the signals on itself as the first thing it does, which still guarantees a second signal mid-shutdown can never land on a worker thread — the original crash stays fixed.

Ran it locally: built clean, started with no Diretta target reachable (so it sits in the "Target not found, retrying..." loop), sent SIGINT — it responded immediately ("Signal 2 received, shutting down...") and exited cleanly (code 0). Previously-flagged regression confirmed fixed.

One minor, non-blocking note: there's a microscopic window between a worker thread's creation and it executing its first instruction (the block call) where it still carries the main thread's unblocked mask — but that only matters at startup thread-creation time, not during shutdown (the scenario this PR actually fixes), so it's not a practical concern.

Merging.

Issue #80, opened 2026-06-29
[FEATURE]  Add DoP (DSD over PCM) output option for DACs requiring DoP


yama3kzh:
## Summary
I own multiple DACs, and one of them produces noise when playing DSD unless the signal is delivered as DoP (DSD over PCM).  
For this DAC, I need to enable the DoP option in other renderers such as HQPlayer or TuneBrowser.

However, diretta_renderer does not provide a DoP output option.  
I would like to request consideration for adding DoP support.

## Background
In my environment:

- Some DACs accept Native DSD
- One DAC requires DoP for clean DSD playback
- Without DoP, DSD results in audible noise

Other renderers (HQPlayer, TuneBrowser, MPD) provide a DoP output option, and enabling it solves the issue.  
diretta_renderer currently outputs PCM only, and DirettaRendererUPnP also does not generate DoP.

As a result, DSD cannot be played correctly on DACs that require DoP.

## Request
Please consider adding a DoP output option to diretta_renderer, such as:

- ALSA output parameter (e.g., `dop "yes"`)
- Automatic DoP framing when receiving DSD input
- PCM fallback when DoP is disabled

## Motivation
Adding DoP support would:

- Enable DSD playback on DACs that require DoP
- Improve compatibility with existing audio systems
- Provide behavior similar to HQPlayer, TuneBrowser, and MPD
- Reduce dependency on specific UPnP servers that generate DoP

Thank you for considering this feature request.

cometdom:
Thanks for the detailed request — DoP output is a valid use case and worth considering.

Before committing to an implementation, a few clarifying questions:

**1. What is your UPnP source / control point?**
For example: Audirvana, JRiver, MinimServer, Roon (UPnP bridge), BubbleUPnP server, etc.

The reason this matters: some UPnP sources already transmit DSD *as DoP* (packed into 24-bit PCM frames with `0x05`/`0xFA` markers), rather than as a native DSD stream (DSF/DFF). If your source already sends DoP, DRUP receives it as plain 24-bit PCM and the Diretta Target passes it through unchanged — in which case it may already work today without any changes, as long as the sample rate (176.4 kHz for DSD64, 352.8 kHz for DSD128) is supported.

**2. What Diretta Target are you using?**
(e.g. GentooPlayer, AudioLinux, Memory Play, a licensed commercial target)

**3. How is your DAC connected to the Diretta Target?**
[...]

PR #90, opened 2026-09-07
Sound-quality pass on the host side: S24/seek/EOF fixes, HTTP prefetch thread, quieter host, SDK options, decode test harness

herisson-88:
> **Informational PR — do not merge as one unit.** Per your review, the content is now split into six PRs rebased on `main`, each with its own CHANGELOG lines; the union of the six is identical to this branch (checked by merging them together and diffing). This one stays open as the whole-picture reference: the review thread, the listening notes and the analysis documents (`docs/analysis/2026-09-sq-review/`, `docs/CLAUDE.md`) live only here.
>
> | PR | Content | Depends on |
> |---|---|---|
> | #91 | Seek: ring flush, request never lost, seek while paused | — |
> | #92 | Tuner drop-ins without process-wide FIFO, doc/conf corrections | — |
> | #93 | `--port-strict` (opt-in fixed port after a hot restart) | — |
> | #94 | SDK usage: `--sink-buffer-ms`, `connect(cpu)`, `auto-sdk`, `--rapid-start`, negotiated profile log, FIX-profile alignment | — |
> | #95 | Correctness fixes: S24 hint, whole-frame pushes, sink negotiation, RT-thread logging, EOF drain, journald flush | — |
> | #96 | Quieter host, HTTP prefetch thread, FLAC bypass, test harness, tools and suites | #95 |

## Summary

A sound-quality pass on the host side of DRUP, done as a branch of v2.5.15 so you get finished, tested commits rather than a list of findings to re-triage. It comes out of a full read of the 2.5.15 tree and of the SDK 149/150 headers and docs (the review itself is included, in French, under `docs/analysis/2026-09-sq-review/`).

Nothing DSD-specific was touched (no DSD hardware to test on). Everything else was built and run on an x86 Fedora PREEMPT_RT host with SDK 150_4 and a Holo Audio Red target, verified bit-exact against reference PCM with a new test harness, and validated by listening — including a live A/B session on the new SDK options that turned up (and fixed) three more issues, see *Live session* below.

**Ten commits, one per theme, each self-contained and bisectable** (every one builds against SDK 150; the first also against SDK 149). Every fix found in review — yours and a second, adversarial pass of my own (last section) — is folded into the commit that made it necessary; each commit message ends with an "After review" paragraph saying what changed in it.

| Commit | Theme | What changes |
|---|---|---|
| `7518299` | `feat:` SDK usage | `setSink()` buffer time (`--sink-buffer-ms`; unset = the cycle time 2.5.15 always passed, `0` = sink default), `Sync::connect()` on the first audio core (overload selected at compile time, SDK 149 keeps building), `--transfer-mode auto-sdk` with FIX-profile buffer alignment on `getCycleSize()` (DoP-even), `--rapid-start`, negotiated profile logged at every `OPEN`. |
| `47dc2db` | `fix:` | S24 hint kept across pause/resume + MSB default, whole-frame ring pushes, `push32To16`, RT-thread events via atomic mailbox, `configureSinkPCM()` returns `bool` (32/24/16 · 24/32/16 · 24/16/32), output refusal tears down properly (no PLAYING zombie), EOF drain sequence, `TimestampedStreambuf` `sync()` + `xsputn()`, four new ring unit tests. |
| `48a88dc` | `perf:` | Decode hysteresis (~20 wake-ups/s), self-pipe shutdown, real `--quiet` (SIGUSR1 stats and lifecycle warnings on stderr), `mallopt` (32 MB threshold, 1 MB pad — locked RSS back to ~100 MB), helper threads demoted and signal-masked, probe cap, hot-path memory ordering, RT mailbox on its own cache line, cadence statistics. |
| `d2b54a7` | `feat:` | `PrefetchReader` (partial reads, low-water refill, copy outside the lock, own interrupt callback, 256 KB cap while preloaded, prompt `stop()`), FLAC on the bit-perfect bypass, `make test-decode`; dead `src/sync/` removed. |
| `bb44739` | `chore:` | Tuner drop-ins without process-wide FIFO; doc/conf corrections (`--cycle-time` examples recomputed, README help block, `SystemCallFilter=` untouched). |
| `74da527` | `test:` | `tools/range_server.py`, `tools/make_test_audio.py` (WAV generator + minimal FLAC writer), `tools/dsf2dff.py` (lossless DSF → DFF for comparing the two DSD parsers). |
| `7086922` | `fix:` | Seek: ring flush, request never lost, seek while paused (cleared on Stop or URI change), unit check. |
| `fff41b3` | `test:` | `tests/decode_suite.sh` (real files, prefetch vs no-prefetch, seeks) and `tests/bitexact_suite.sh` (synthetic files, hash vs source samples). |
| `a343dee` | `fix:` | Port after a hot restart: `--port-strict` / `PORT_STRICT=1` waits for the configured port (interruptible, stderr messages) for control points that cache the address (JPLAY); default = libupnp's port+1 behaviour with a warning naming the port used. |
| `0463758` | `docs:` | `docs/CLAUDE.md`, CHANGELOG, review documents, `.gitignore` for audio. |

## Bugs fixed

- **24-bit alignment (pause → resume).** the ring's sample-sniffing S24 detection timed out into `LsbAligned` after ~1 s of silence, and `clear()` — called on pause → resume — forgot the alignment hint while nothing re-set it (resume does not go through `open()`). Pause, resume while the music is quiet for more than a second, and the ring locked to LSB: full-scale noise once the music came back, on x86 (ARM forced MSB unconditionally, hiding it). A fresh track was not affected — the hint was already set for practically every codec at open. Fix: the hint survives `clear()` (only `resize()` forgets it and its caller re-sets it), the timeout/fallback default is MSB (the only alignment FFmpeg produces), the `#if __aarch64__` special case is gone, and the hint is set for every 24-bit track.
- **Frame-aligned ring pushes (latent).** `getFreeSpace()` is `size − used − 1` and the PCM `push*()` clamped to bytes/samples rather than frames, so a push into an almost-full ring could write a partial frame. The producer advances by the bytes actually written, so the byte stream stayed continuous; the damage was in its sample accounting (floor division), which could make the last `sendAudio()` of a callback read up to one frame past the callback buffer and push a garbage frame. With the 50 % throttle the ring never got that full, so this was essentially unreachable in normal operation. Pushes are whole-frame now regardless.
- **EOF drain (hygiene, not a measured fix).** `readSamples()` now sends the flush packet, flushes the resampler and serves the FIFO after EOF — FFmpeg's documented end-of-stream sequence. Measured against v2.5.15 on FLAC 16/44, FLAC 24/192, ALAC 24/96 and MP3: frame counts and output hashes are identical, i.e. v2.5.15 was not losing frames on these codecs (they deliver one frame per packet). Kept as correctness-by-contract for codecs that do buffer internally; no measured case where it changes the output.
- **Sink negotiation.** `configureSinkPCM()` threw `std::runtime_error` through `open()` (an unsupported format took the process down); it returns `false` and the track fails cleanly. Orders: 32-bit source 32 → 24 → 16, 24-bit source 24 → 32 → 16, 16-bit source 24 → 16 → 32 — 32-bit is never offered first to a 16/24-bit source (v2.4.4, TEAC UD-701N), it is only a last resort on a sink that refuses 24-bit, where it is lossless for us (the ring holds S24 in S32). A 24/32-bit source on a 16-bit-only sink used to `memcpy` 4-byte samples as 2-byte ones; it is now an explicit MSB truncation (`push32To16`). DoP requires an exact 24-bit sink.
- **iostream on the SDK real-time thread.** `getNewStream()` did `LOG_WARN` (mutex + `write(2)` to journald) on underrun and on rebuffering complete. It now raises bits in an atomic mailbox that the decode thread logs.
- **Seek.** Nothing flushed the Diretta ring after a seek: everything buffered from the old position (up to 0.5 s local / 3 s remote) played first while `GetPositionInfo` already reported the new position — repeated seeks queued seconds of stale audio. The ring is now emptied under the reconfigure guard and the prefill restarts. The async request was consumed as "read target, then clear flag", which drops a seek that lands in between (scrubbing control points send several per second) — now `exchange()` first. A seek while paused was refused; it is queued and applied on resume, cleared on Stop. `Seek` with unit `TRACK_NR` was applied as seconds; ignored.
- **Preload thread scheduling.** `std::thread` inherits its creator's policy and affinity; the preload thread was spawned from `process()`, i.e. it ran a 300–500 ms HTTP open + probe at `SCHED_FIFO` on the decode core. Demoted to `SCHED_OTHER` on the `--cpu-other` cores.
- **Log lines never reaching journald (pre-existing).** `TimestampedStreambuf` did not override `sync()`, so `std::endl` never flushed the stdio buffer behind `std::cout`. On a terminal (line-buffered) nothing shows; under systemd stdout is a socket, fully buffered, and lines only surfaced once 4 KB had piled up — which is why the renderer's log always looked "late" in the journal, and why a quieter renderer showed nothing at all.
- **Port drift after a hot restart (pre-existing, opt-in mitigation).** libupnp is built without `SO_REUSEADDR` on its miniserver. Restart while the control point still has connections to the old instance: they sit in `TIME_WAIT` on the configured port, `bind()` fails, libupnp silently takes port+1 and announces it over SSDP. Most control points follow; JPLAY keeps its cached LOCATION and never reaches the new instance (seen twice in one session). `--port-strict` (`PORT_STRICT=1`) makes `UPnPDevice::start()` retry every 2 s until it gets the configured port (up to 75 s, interruptible; verified: "Port 4005 busy … libupnp offered 4006 — waiting" → "Port 4005 acquired after 54s"). It is opt-in because it costs up to a minute without a renderer after every hot restart; the default keeps libupnp's behaviour and logs a warning naming the port actually used.
- **Docs.** `--thread-mode` bit 8 is not `SOCKETNOBLOCK` (commented out in the SDK header), `FEEDBACKOFFSET` is a 3-bit field, `--cycle-time` default/range, remote buffer defaults (3.0 s / 500 ms in the code, 1.0 s / 150 ms in the conf and README).

## Host activity

- **`PrefetchReader`** — every HTTP source is read by a dedicated `SCHED_OTHER` thread on the `--cpu-other` cores into a 4 MB byte ring; the demuxer reads from that ring through a custom `AVIOContext` (`AVFMT_FLAG_CUSTOM_IO`, same lifetime handling as the existing Audirvana wrapper). Seeks are forwarded to the inner HTTP context as `Range` requests after draining the ring; the decoder's stall interrupt callback keeps working. The decode thread — `SCHED_FIFO` on its own core — no longer performs any network I/O. `--no-prefetch` / `NO_PREFETCH=1` restores the previous behaviour for A/B.
- **FLAC on the bypass** — `canBypass()` refused every "compressed" format on the assumption that they decode planar. FLAC emits packed S16/S32 (S24 left-justified in S32, byte-identical to what `swr_convert` produced); it now takes the bypass. ALAC/WavPack are planar and stay on the resampler.
- **Decode thread hysteresis** — fill the ring to 70 %, sleep (5–50 ms, sized from the measured drain rate) until 30 %, fill again: a few decode bursts per second instead of a 10 ms poll and a 46 ms chunk as soon as the level dips under 50 %. In the live session the ring sat at 66–67 % during playback, 0 underruns.
- **`--quiet` for real** — ~250 raw `std::cout` sit on the track path and each `std::endl` was a `write(2)` to journald from whichever thread emitted it. Quiet mode swaps `std::cout`'s buffer for a null sink; warnings and errors go to `std::cerr` and are untouched.
- **Stable heap** — `mallopt(M_MMAP_THRESHOLD=1 GB, M_TRIM_THRESHOLD=-1, M_TOP_PAD=64 MB)` before `mlockall()`: the heap never shrinks nor mmaps per block, so track open/close no longer costs `mmap`/`munmap` and fresh page faults.
- Removed the empty 1 Hz "UPnP Thread" and the main thread's 1 Hz `isRunning()` poll (the main thread now blocks in `read()` on a self-pipe written by the signal handler — the handler no longer calls `stop()`/`exit()` itself, see *Live session*). FFmpeg probe capped for every URL (32 KB local, 256 KB / 1 s remote — was the 5 MB default on remote streams).
- `getNewStream()`: `m_workerActive` stores are `release` (were seq_cst, an `mfence` each), the stream counter is a plain store, the flow-control condvar is only touched when a DSD producer waits, worker-only cached state on its own cache line. Call-interval statistics (mean/min/max/late vs expected cycle) in `dumpStats()` (SIGUSR1).
- **systemd** — the tuner drop-ins set `CPUSchedulingPolicy=fifo` / `Priority=90` on the whole process, i.e. every control thread above the audio worker at `RT_PRIORITY`; removed (the renderer sets FIFO per thread itself). `SystemCallFilter=` stays (an earlier revision commented it out; restored after review — the seccomp cost is tens of nanoseconds per syscall, not worth dropping hardening).

## How it was tested

**Bit-exactness.** `make test-decode` builds `bin/test_decode`, which runs a URL through `AudioDecoder` exactly like the renderer does (prefetch thread, bypass, EOF drain, seeks) and prints the frame count and an FNV-1a hash of the output bytes. `tests/decode_suite.sh <dir>` serves a directory with `tools/range_server.py` and checks, for every file:

1. full decode with the prefetch thread → frame count == container duration;
2. full decode with `--no-prefetch` → same hash;
3. five successive seeks (25 %, 75 %, 5 %, 90 %, 40 % of the duration) in both modes → same hash, frames ≈ expected;
4. thirty random rapid seeks → no error, frames ≈ expected.

Results on real files — FLAC 16/44 (325 s), FLAC 24/192 (179 s, with SEEKTABLE), ALAC 24/96 (39 s), MP3 44.1 k (112 s): **20/20**, also under AddressSanitizer. **Against v2.5.15**: the original decoder built with the same harness gives *identical* frame counts and *identical* output hashes on all four files — the branch's decode-path changes (FLAC on the bypass instead of `swr_convert()`, EOF drain, S24 hint handling) are output-neutral on real material. That is the intended result: bit-perfect stays bit-perfect; the gains of the branch are in host activity, seek behaviour and robustness, not in the samples. On synthetic WAV/FLAC material the hashes were also compared to the source PCM itself (24-bit expanded to S32 as FFmpeg does): identical.

A note for anyone reproducing: `python3 -m http.server` ignores `Range`, and FFmpeg's generic FLAC seek bisects with backward seeks — every FLAC seek then fails with "read_timestamp() failed in the middle", on v2.5.15 as well. It is not a renderer bug; use `tools/range_server.py` (or any real server).

**Listening.** Deployed on a fanless i5 host (Fedora PREEMPT_RT, `nosmt isolcpus=1-3`, `--cpu-audio 2 --cpu-decode 3 --cpu-other 1`, FIFO 80) driving a Holo Audio Red, control point JPLAY iOS, sources Qobuz via JPLAY (16/44, 24/44.1, 24/48, 24/96 — all on the bit-perfect bypass, 0 underruns) and local FLAC. Audible improvement over 2.5.15 on the same chain; seeks (single, repeated, while paused) behave as expected.

## Live session on the new SDK options

Each option was switched on with one clean restart, then judged by ear and by the profile logged at `OPEN`. Baseline on this chain: `cycle=14445µs cycleSize=3822B (MTU 3824) packets/cycle=1 mode=VARIABLE msMode=MS3 latency=110ms`.

| Option | Result |
|---|---|
| `--transfer-mode auto-sdk` | **Silent at first, fixed, then kept.** `configTransferAuto()` negotiated `mode=FIX cycle=2000µs cycleSize=704B` while the renderer handed the SDK its usual 1 ms buffers (352 B at 16/44): two half-cycles per fixed cycle. The ring filled normally (66 %, 66 k `getNewStream()` calls, 0 underruns), the target stayed silent. In VARIABLE mode the SDK adapts to whatever we give; a FIX profile sends exactly `getCycleSize()` per cycle and expects that much per callback. `7518299` aligns the callback buffer on the negotiated cycle size after `applyTransferMode()`/`connectWait()` (`FIX profile: callback buffer 352 -> 704 bytes (88 frames)`), drift accumulator off in that mode. After that: 16/44, 24/44.1 and 24/96 play (cycle sizes 704/528/1152 B, i.e. 88/88/192 frames), cadence measured 1997 µs mean for 2000 µs expected, max 4000 µs, 1 late interval in 3582, 0 underruns. Listened to for a while, then **dropped again**: the tester preferred the VARIABLE baseline (14 ms MTU-sized cycles) to the 2 ms fixed cycles on this chain. So: it works now, it is a real alternative regime to compare on other targets, and it stays off by default. |
| `--sink-buffer-ms 100` (the SDK sample host's value) | Works, mode and cycle unchanged, and the `latency` the SDK reports is 110 ms with **or** without it (it comes from the sink's own `latencyBuffer=1000`), so the effect is on the target side only. Kept for a while, then dropped after longer listening. Note: an earlier revision of the branch had silently changed the *default* from "the cycle time" (what 2.5.15 always passed) to 0; the default is 2.5.15's value again, and the tester's listening was done with `0`, now pinned explicitly in his config. |
| `--rapid-start` | No measurable change on the host: full `OPEN → OPEN COMPLETE` takes 0.55–0.87 s without it (7 tracks) and 0.52–0.86 s with it (6 tracks). The tester first reported a snappier start, then dropped it after longer listening. If the effect is real it is in the target's start-up after connect, which the SDK does not expose. Harmless either way; off by default. |
| `--no-prefetch` | Not A/B'd by ear yet (bit-exact equivalence is covered by the suite). |

With a FIX profile the "late" counter of `dumpStats()` becomes meaningful (one callback per cycle) — the 1997/2000 µs figure above comes from it.

What the session also produced (now folded into their theme commits): the branch's own shutdown deadlock — main thread blocked on a condition variable that the signal handler's `exit()` then tried to destroy, 45 s systemd timeout + SIGABRT on every restart — and the pre-existing flush bug that hid all renderer output from journald) and the port-drift fix (`a343dee`). One caveat on my own instrumentation: the "late (>2× expected)" counter in `dumpStats()` compares intervals against *our* buffer time; in VARIABLE mode the SDK drains a whole cycle of 1 ms buffers back-to-back (13 calls, then a 13 ms gap), so half the intervals are 0 µs and half are a full cycle. The counter should be re-based on the SDK cycle (`getCycleTime()`); it is informational only.

**DSD, decoder level.** Free native DSD64 and DSD128 DSF recordings (HifiStatement, Godard *A Trace of Grace*) decode through the prefetch thread with identical output with and without it; the decoder returns exactly the DSF header's sample count (FFmpeg's duration estimate includes the last block's padding, the suite tolerates that); a DFF produced from the DSD64 DSF with `tools/dsf2dff.py` goes through the built-in DSDIFF parser and yields, bit for bit, the bit-reversed DSF output; a 30 s seek gives identical results in both modes. **DSD on the target.** The same DSD64 DSF, DSD128 DSF and DSD64 DFF were played on the Holo Audio Red through the live renderer (default transfer mode): DSD negotiated (`2822400Hz/1bit` and `5644800Hz/1bit`), 0 underruns over 45 s each, ring 33–66 %, cadence 997/1003 µs measured for 997/1003 expected, no warnings — so the DSF and DFF paths through `pushDSDPlanarOptimized` (with the DFF bit reversal) work end to end. **DoP on the target** (Red → Holo Spring 2 KTE): `DOP=1` on a DSF gives music plus constant hiss, `DOP=msb` plays identically to native DSD — the same conclusion issue #80 reached on the SFORZATO, since DoP v1.1 wants the oldest sample in the MSB and DSF stores it in the LSB. Behaviour left as is; noted here for the docs. **Not tested:** `alignBufferToNegotiatedCycle()`'s DoP parity rule, only exercised with `--transfer-mode auto-sdk/fixauto`. ARM builds (the `#if __aarch64__` removal is intentional and the replacement default is the one ARM was already forcing). Other DACs than the Red for the 24→32 sink fallback (it only triggers on a sink that refuses 24-bit).

## Second review pass (adversarial, four angles, findings verified on the host)

Run after the first review round: correctness/concurrency, hot path, removed behaviour vs the project's history, reuse/conventions/tests. Everything below is folded into the theme commits.

**Real and fixed**
- `mallopt(M_MMAP_THRESHOLD, 1 GB)` was silently *rejected* by glibc (max 32 MB) while `M_TOP_PAD = 64 MB` — per arena, locked and populated under `MCL_FUTURE` — took the locked RSS from ~100 MB (2.5.15, 23 h run) to 600 MB–1 GB (journal `memory peak` of every branch instance). Now 32 MB (return checked) / 1 MB: idle RSS measured **101 MB** after the fix.
- `PrefetchReader` did a full 256 KB `avio_read()` before forwarding anything: 16 s for the first byte on a 128 kbit/s radio, i.e. the v2.5.11 stall deadline fired — `avio_read_partial()` now. A preloaded next track pulled its whole 4 MB ring while the current track played (the concurrent-read corruption e4c4428 exists for) — capped at 256 KB until the track becomes current. The reader could outlive a format context whose interrupt callback it chained to (every `open()` failure after `avformat_open_input`) — those exits go through `close()`. Copy into the ring outside the mutex, low-water refill, EOF/error classification from the inner context, failed Range seek no longer leaves the reader "stopping".
- Sink negotiation failure left the UPnP state machine in PLAYING with no way forward — tears down like a fatal decode error now.
- The RT event mailbox shared a cache line with `m_streamCount`/`m_prefillComplete` and the decode thread did a locked `exchange` on it every iteration — own line, relaxed poll.
- `setSink()`: 2.5.15 passed the cycle time, not 0; the branch's default had changed it silently. Default restored; documented.
- A seek queued while paused survived a URI change; DoP even-frame invariant in FIX alignment; `LOG_WARN` on unbuffered stderr wrote one byte per character (no `xsputn`); the `--quiet` null sink was a function-local static destroyed before `ios_base::Init`; the hysteresis comment claimed 2–3 wake-ups per cycle where arithmetic gives ~8 (sleep now sized to LOW, cap 100 ms, ~20/s); doc contradictions (SystemCallFilter, condition variable, `--cycle-time` examples off by ~3×, README help block, v2.4.4 item); no ring unit tests for the S24/frame/32→16 changes (four added); the bit-exact claim was a manual step (now `tests/bitexact_suite.sh`).

**Reviewed and rejected (with evidence)**
- "EOF drain fixes lost tails": 2.5.15 produces identical frame counts and hashes on FLAC/ALAC/MP3 — reclassified as hygiene.
- "Torn frames → channel swap": latent and mis-described — corrected.

**Known and left**
- `src/test_audio_memory.cpp`: the two DoP tests (`pushDSDToDoP` byte layout, MSB bit reversal) fail on v2.5.15 as well — they were not updated for the v2.5.10 byte-pair fix. Not touched by this branch.
- The fixed-port wait (`--port-strict`, opt-in) is a workaround for libupnp built without `SO_REUSEADDR`; the root fix is `--enable-reuseaddr` at package/build level.
- Removing the process-wide `CPUSchedulingPolicy=fifo` from the tuner drop-in means SDK-internal threads other than the `getNewStream()` caller (if any) are no longer FIFO 90 — no measurement either way; worth an A/B on other targets.

## A/B knobs

All default to the previous behaviour: `NO_PREFETCH=1` (synchronous reads on the decode thread, as before), `SINK_BUFFER_MS` (default 0 = sink default, as before; the SDK sample uses 100), `TRANSFER_MODE=auto-sdk` (the SDK sample host's mode; FIX 2 ms cycles on a Holo Red), `RAPID_START=1`. After the session the tester went back to the branch defaults — no sink buffer override, VARIABLE transfer mode, no rapid start: none of the three SDK options survived a longer listen on this chain. They remain available for other targets. The hysteresis thresholds, `--quiet` semantics and `mallopt` are unconditional.

## Rollback

Every commit is self-contained. The systemd changes are two lines in the drop-in and one in the unit; `install.sh` users get them with the next install, existing installs need the drop-in regenerated by the tuner.


herisson-88:
Listening verdict after a full evening on the branch (Holo Audio Red, JPLAY iOS, Qobuz 16/44 → 24/96 and local FLAC):

**The branch defaults are clearly better than 2.5.15 on this chain** — that part is not in doubt, and it is what is running now: `--quiet`, VARIABLE transfer mode, sink buffer left to the target, nothing else.

**The new SDK-150-era options are doubtful**, and I would not recommend them from this test:

- `--transfer-mode auto-sdk` — works now (FIX 2 ms cycles, buffer aligned on `getCycleSize()`), plays every format cleanly with 0 underruns, but sounded *worse* than the VARIABLE baseline. Kept in the code because it is the SDK sample host's mode and other targets may behave differently; off by default.
- `--rapid-start` — no measurable effect on the host side (`OPEN → OPEN COMPLETE` 0.52–0.86 s either way); the "snappier start" impression did not survive longer listening. Off by default.
- `--sink-buffer-ms 100` — the SDK sample's value; harmless, but the target's own default was preferred by ear. Off by default.

So the SDK options are best read as *exposed and testable* rather than *recommended*. The value of this branch is in the fixes (S24, torn frames, decoder tail, seek, shutdown, journald flush, port drift), the prefetch thread, the FLAC bypass and the quieter host — all of which are on by default and all of which were on during the listening that beat 2.5.15.

cometdom:
Thanks for the depth here — the SDK-header read, the bit-exact test harness, and the live A/B session on the new SDK options (catching the shutdown deadlock and port-drift bugs along the way) are genuinely excellent work. Reviewed at high effort across several angles (correctness, hot-path efficiency, reuse, root-cause-vs-bandaid, removed-behavior, conventions) and verified the most serious findings myself, not just from static analysis. Given the scope (11 independent commits, ~2500 lines, hot path + SDK + systemd), I don't think this is mergeable as one unit yet — several real bugs, some of them serious. Since every commit is explicitly self-contained per your own summary, cherry-picking the clean ones (tests, tooling, most of the docs/deployment commit) while holding back the ones below seems like the practical path.

## Verified critical issues

**1. Breaks the build against SDK v149** (`f89e0ed`, `src/DirettaSync.cpp:845`)
```cpp
connected = connect(sdkConnectCpu, m_config.rapidStart);
```
SDK v149's `Sync::connect(int)` takes one argument; the two-arg overload is v150-only. I compiled this branch against `DirettaHostSDK_149` to confirm — it fails immediately:
```
src/DirettaSync.cpp:845:32: error: no matching function for call to 'DirettaSync::connect(int&, bool&)'
```
CLAUDE.md documents (and a previous session verified) that v149 still builds and runs fine — this regresses that. Needs an SDK-version guard (`#if` on `DIRETTA::ReleaseNo` or similar) around the `rapidStart` argument.

**2. Likely heap overflow in the new EOF-drain path** (`6221e16`, `src/AudioEngine.cpp:1580`)
```cpp
size_t samplesNeeded = numSamples - totalSamplesRead;
```
Once `draining` is true, the receive loop (`while (ret >= 0 && (draining || totalSamplesRead < numSamples))`, line 1525) deliberately keeps pulling frames past `totalSamplesRead == numSamples` to drain codec lookahead (ALAC/AAC/MP3/Vorbis/WavPack). Once `totalSamplesRead > numSamples`, this subtraction underflows (unsigned) to a huge value, so `samplesToCopy = std::min(frameSamples, samplesNeeded)` (lines 1607/1693) and `samplesToUse` (line 1667) stop being clamped by remaining buffer space — the full `frameSamples` gets copied via `memcpy_audio()` into `outputPtr`, which is already at or past the buffer's original allocated end. Reachable on essentially any FLAC/ALAC/AAC/MP3/WavPack track's EOF where the decoder yields more than one frame after the flush packet — not a rare codepath.

**3. Regresses the v2.4.4 bit-depth fix** (`6221e16`, `src/DirettaSync.cpp` ~1086-1090)
The new `configureSinkPCM()` fallback order lets 16-bit and 24-bit sources negotiate a 32-bit sink (24-bit source tries `{32,24,16}`, 16-bit tries `{32,24,16}` per the PR description). CLAUDE.md's Bit Depth Handling section is explicit: *"16-bit and 24-bit sources: Only negotiate up to 24-bit"* — added specifically because DACs like the TEAC UD-701N advertise 32-bit at the Diretta target level while being physically 24-bit-limited (v2.4.4/v2.4.5, silence/noise symptom). This reopens that exact bug for any 16/24-bit source on such a DAC.

**4. `PrefetchReader`'s thread doesn't block shutdown signals** (`7b7d480`, `src/PrefetchReader.cpp`)
Every other worker thread (`upnpThreadFunc`, `audioThreadFunc`, `positionThreadFunc`, `g_logDrainThread`) calls `blockShutdownSignalsOnThisThread()` as the first thing it does, specifically because of the v2.5.13 fix: a second SIGINT/SIGTERM landing on any thread that doesn't block it can re-enter `signalHandler()` concurrently with an in-progress shutdown and crash with `terminate called without an active exception`. The new prefetch thread is a fresh, unguarded target for that exact race, and it's on by default.

**5. `PrefetchReader::stopThread()` can hang indefinitely** (`7b7d480`, `src/PrefetchReader.cpp`)
`stopThread()` does a bare `m_thread.join()` with no timeout. The reader thread's blocking `avio_read()` is only interruptible via `AudioDecoder::m_readDeadlineNs`, which is armed transiently around the *decode* thread's own `av_read_frame()` calls and cleared right after — it's not armed for the independent network I/O now happening on the prefetch thread. If that thread stalls mid-`avio_read()` (the exact proxy-keeps-TCP-alive-no-data scenario v2.5.1/v2.5.11 exist for) and something calls `stop()`/`close()` — a UPnP Stop, a URI change discarding a stale preload, or the existing reconnect path itself — `stopThread()`'s join can block the calling thread (audio thread or a UPnP callback thread) indefinitely. Same hang class those earlier releases closed, one layer deeper.

**6. Data race on the new stats counters** (`be69f23`, `src/DirettaSync.h`/`.cpp`)
`m_streamIntervalMinNs/MaxNs/SumNs/Count/Late` are plain (non-atomic) members written by `getNewStream()` on the SDK's real-time worker thread and read by `dumpStats()`, reachable from a different thread via the `SIGUSR1` handler — a data race under the C++ memory model, at minimum producing torn/inconsistent values.

**7. `SystemCallFilter=` hardening removed with no replacement** (`261c9c8`, `systemd/diretta-renderer.service`)
The seccomp syscall filter (`~@mount @keyring @debug @module @swap @reboot @obsolete`) is commented out entirely — every other hardening directive stays. The service runs as root; without this, a compromised process (e.g. via a malicious/corrupt stream hitting a decode bug) could call `mount()`/`keyctl()`/`swapon()`/`reboot()`/load kernel modules where it previously couldn't. If the BPF overhead on `sendto()`/`recvfrom()` is the real concern, a scoped filter would be safer than none.

## Worth addressing too

- **New UPnP port-retry loop is uninterruptible for up to 75s** (`9c3c2e1`, `src/UPnPDevice.cpp`) — plain `sleep_for(2s)` loop with no `stopSignal` check, unlike the pattern `DirettaRenderer::start()` already uses (and v2.5.13 specifically hardened) for the sibling "Network not ready, retrying..." loop. Root cause is likely libupnp's miniserver built without `SO_REUSEADDR` — patching/rebuilding libupnp (the project already builds FFmpeg from source for similar reasons) would avoid the wait entirely rather than riding out `TIME_WAIT`.
- **Likely doubled timestamps in verbose logs** — the new `TimestampedStreambuf` prepends a wall-clock prefix to every `std::cout` line process-wide, but `logDrainThreadFunc()` already prints its own `[Nms]` prefix per drained entry (`src/main.cpp`) — these look additive now (`[14:23:01.123] [45912ms] message`).
- `canBypass()`'s blanket "compressed formats never bypass" guard was narrowed to a format-shape check (packed vs planar) — if any lossless codec besides FLAC ever decodes to non-planar S16/S32, it'd now silently take the raw-copy path instead of going through `swr_convert()`.
- The Qobuz/Tidal preload-probe fix stays keyed on a hardcoded provider-name string match rather than "any HTTP URL", per your own review doc's stated general recommendation — a third relayed service (or an obfuscated proxy URL for the same two) falls through unprotected.
- `PCM_REMOTE_BUFFER_SECONDS`/`PREFILL_MS` raised (1.0s→3.0s / 150ms→500ms) in the same PR that adds `PrefetchReader` specifically to smooth the jitter those margins were compensating for — worth re-measuring against the new prefetch thread rather than keeping both the fix and the inflated buffer.
- Minor: `push32To16()` is scalar while every sibling conversion (24-bit pack, 16→32) is AVX2/NEON; `PrefetchReader` reimplements a mutex+condvar ring instead of reusing `DirettaRingBuffer`'s lock-free SPSC; CPU-affinity-setting is now a 3rd near-identical copy (`demoteToHelperThread` vs `pinThreadToCores`/`pinCurrentThread`); a few unused members/accessors (`m_threadRunning`, `pb()`, `buffered()`); two lines over the 120-char limit; several commit subjects don't use the `type: description` convention from CLAUDE.md.

Given every commit is independently cherry-pickable per your summary, I'd suggest holding back `f89e0ed`, `6221e16`, `7b7d480`, `be69f23`, and `261c9c8`'s `SystemCallFilter` line pending fixes, while the tests/tooling/docs commits and the seek fix (`394613f`) look solid on their own — happy to help split this up if useful.

herisson-88:
Thanks for the thorough review — most of it is right and it is all addressed in `14df0d6` (built against SDK 149 *and* 150, decode suite 20/20 normally and 20/20 under AddressSanitizer). Point by point, with the two places where I think the analysis was off:

**1. SDK 149 build break — correct, fixed.** `Sync::connect()` now goes through a small `sdkConnect()` helper that picks the one-argument (≤149) or two-argument (150, `rapidStart`) overload at compile time (SFINAE on the overload, no version macro to maintain); `--rapid-start` logs a warning and is ignored on 149. Verified by building the tree against `DirettaHostSDK_149` and `_150`.

**2. Heap overflow in the EOF drain — I don't think this one is real, but it is now provably not.** Every copy in the receive loop is clamped by `samplesNeeded` (`min(frameSamples, samplesNeeded)` / `min(convertedSamples, samplesNeeded)`), so `totalSamplesRead` can reach `numSamples` but never exceed it: once it does, `samplesNeeded == 0`, nothing is copied to `outputPtr`, and the whole frame goes to the FIFO — which is exactly what the draining loop is for. The subtraction therefore never underflows. To settle it rather than argue it, `tests/decode_suite.sh` was run with `AudioEngine.cpp`/`PrefetchReader.cpp` built under `-fsanitize=address` on FLAC 16/44, FLAC 24/192, ALAC 24/96 and MP3 (ALAC and MP3 flush several frames at EOF): 20/20, no ASan report. I still added an explicit `(totalSamplesRead < numSamples) ? … : 0` guard so the invariant is visible at the point of use.

**3. Bit-depth regression — half right, fixed.** For 24-bit sources the order was already 24 → 32 → 16, i.e. a DAC that announces 32 but is physically 24 negotiates 24 exactly as in v2.4.4; 32 is only reached on a sink that *refuses* 24, where it is lossless for us and beats truncation. For 16-bit sources I had put 32 first — that was wrong and would reopen the TEAC case; it is back to 24 → 16, with 32 as a last resort. CLAUDE.md's "Bit Depth Handling" section now states the three orders.

**4. Prefetch thread and shutdown signals — correct, fixed.** `demoteToHelperThread()` (used by the prefetch *and* the preload thread) now blocks SIGINT/SIGTERM first thing. Note that since `4d460d9` the handler itself no longer calls `stop()`/`exit()` — it writes one byte to a self-pipe and the main thread does the shutdown — so a signal landing on a worker thread could no longer re-enter a shutdown anyway; the mask keeps the v2.5.13 rule intact regardless.

**5. `stopThread()` hang — correct, fixed.** The inner HTTP context is now opened with the reader's own `AVIOInterruptCB`, which returns "abort" as soon as a stop is requested and otherwise chains to the decoder's stall-deadline callback. A `stop()` from the audio thread or a UPnP callback returns as soon as the blocked `avio_read()` is interrupted, and the context's sticky `eof_reached`/`error` are cleared before a seek restarts the reader (that one bit the suite immediately — "Seek failed: End of file" on a seek right after an abort — and is covered by it now).

**6. Stats data race — correct, fixed.** The five counters are `std::atomic<int64_t>` with relaxed loads/stores (plain moves on x86/ARM64, no RMW on the hot path); the last-timestamp stays a worker-private plain member.

**7. `SystemCallFilter=` — you're right, restored.** The BPF program costs tens of nanoseconds per syscall; my review doc listed it as a hot-path item without quantifying it, and dropping hardening for that is not a good trade. The unit ships with the filter; the comment says to comment it out only for an A/B.

**Port-retry loop** — now checks the stop signal every 100 ms (`UPnPDevice::setStopSignal()` wired from `DirettaRenderer::start()`), so a shutdown during the wait returns immediately. Agreed that rebuilding libupnp with `--enable-reuseaddr` is the real fix; that is a packaging decision (Fedora's `libupnp` is built without it), the retry is what works with the distro package.

**Smaller points:**
- *Doubled timestamps in verbose logs* — `TimestampedStreambuf` is not new (it dates from v2.0.0); the `[Nms]` prefix of the drain thread has been printed inside a wall-clock-stamped line ever since. What is new is only the `sync()` override that makes the lines reach journald at all. Happy to drop the inner prefix if you prefer, it is one line.
- *`canBypass()` narrowing* — the check is on the actual output sample format (packed `S16`/`S32` matching the requested width), plus the existing runtime guard in `readSamples()` that falls back to the resampler on a frame-format mismatch. Any decoder producing packed S16/S32 yields byte-for-byte what `swr_convert()` would (S32→S32 is a copy), so a future packed lossless codec taking the bypass would be correct, not a silent divergence.
- *Qobuz/Tidal probe keyed on provider strings* — the probe cap now applies to **every** URL (32 KB / 0 s for local servers, 256 KB / 1 s for everything else); the provider-string match only selects the remote HTTP option set, as before.
- *Remote buffer 1.0 → 3.0 s* — not raised by this branch: `PCM_REMOTE_BUFFER_SECONDS = 3.0f` / `PCM_REMOTE_PREFILL_MS = 500` are what `DirettaSync.h` has contained since before 2.5.15; the conf file, README and CLAUDE.md said 1.0 s / 150 ms and were corrected to match the code. No buffer was changed.
- *`push32To16` scalar* — it is the "24/32-bit source on a 16-bit-only sink" path, which previously memcpy'd garbage; a scalar loop is deliberately the minimal correct version. AVX2/NEON can follow if anyone actually hits that path.
- *`PrefetchReader` ring vs `DirettaRingBuffer`* — the lock-free SPSC ring has no blocking wait, and the reader thread *should* block (on the condvar) when the ring is full and when the demuxer wants data that is not there yet; a byte ring with a mutex+condvar on a non-real-time thread is the right tool there. Affinity-setting duplication and the unused members: unused members removed; the three affinity helpers are a small refactor I'd rather do as its own commit if you want it.
- *Commit subjects* — they are on the fork branch of a PR, so I can rewrite them to `type: description` and force-push if you'd rather have that history; the last commit already follows the convention.

On splitting: agreed in principle. With `14df0d6` on top, `6221e16`, `7b7d480` and `be69f23` are what the listening verdict was based on (the branch defaults, no SDK options), so I'd suggest taking them together with the fix commit rather than holding them back — but it is your call, and I'm happy to reorder into whatever grouping you want to merge.

herisson-88:
Self-review pass — I went back over every claim in the description against the v2.5.15 tree and, where it could be measured, against the original decoder rebuilt with the same harness. Three statements were overstated or wrong and are now corrected in the description, `CHANGELOG.md` and `docs/CLAUDE.md`:

1. **"Lost decoder tail" — not a measured bug.** v2.5.15's decoder, built with `test_decode`, produces *identical* frame counts **and identical output hashes** to the branch on FLAC 16/44, FLAC 24/192, ALAC 24/96 and MP3. FFmpeg delivers one frame per packet for these codecs, so nothing was being lost. The EOF flush sequence is kept because it is the documented API contract (and it is what makes codecs that *do* buffer internally correct), but it is filed under hygiene now, not under fixes.

2. **"24-bit alignment: white noise on a quiet track start" — wrong trigger.** At track open the hint was already set for practically every codec (PCM_S24, FLAC/ALAC, any S32/S32P decoder, lossy), so a fresh track was fine. The reachable path was **pause → resume**: `resumePlayback()` calls `clear()`, which forgot the hint, and resume does not go through `open()` so nothing re-set it; resume during a quiet passage longer than ~1 s then locked the ring to LSB and the music came back as full-scale noise on x86 (ARM forced MSB unconditionally). The fix is the same, the description of the symptom is now accurate.

3. **"Torn frames → channel swap" — wrong consequence, and latent.** A partial-frame push could happen when the ring was almost full, but the producer advances by the bytes actually written, so the byte stream stayed continuous; the damage was in the producer's sample accounting (floor division), which could make the last `sendAudio()` of a callback read up to one frame past the callback buffer and push a garbage frame. With the 50 % throttle the ring never got that full, so it was essentially unreachable. Pushes are whole-frame now regardless.

What the same comparison confirms: the FLAC bypass produces exactly the bytes `swr_convert()` produced (identical hashes on both FLAC files), and the whole decode path of the branch is output-neutral on real material. The value of the branch is therefore where the description says it is — host activity, seek, shutdown/journald/port robustness, FIX-profile support — not in the samples, and I'd rather the description say that plainly than claim fixes that don't show up in a measurement.

Two small follow-ups in the same push: `dumpStats()` (SIGUSR1) and the fixed-port wait messages now go to stderr so they remain visible under `--quiet`.

herisson-88:
History rewritten (force-pushed) so that every fix lives in the commit that made it necessary, instead of trailing it five commits later — the branch is now ten self-contained theme commits, `type: description` subjects, and I built each one in turn against SDK 150 (the first also against SDK 149): all ten compile, so it bisects cleanly and cherry-picks by theme. The resulting tree is byte-identical to the previous head.

What moved where:
- shutdown self-pipe, stats atomics, `dumpStats()` on stderr, helper-thread signal mask → `perf:` (the commit that introduced the condition-variable wait and the stats);
- `sdkConnect()` overload selection (SDK 149 build), FIX-profile buffer alignment → `sdk:`;
- 16-bit sink order, EOF-drain guard, `TimestampedStreambuf::sync()` → `fix:` (correctness);
- prefetch interrupt callback / prompt `stop()` / unused members → `feat:` (prefetch);
- `SystemCallFilter=` is simply no longer touched by the `chore:` commit;
- interruptible port wait + stderr messages → the port-drift `fix:`;
- CHANGELOG/CLAUDE.md corrections → `docs:` (last).

Commit hashes quoted in the earlier comments refer to the old history; the description's table is updated. Nothing changed in the code since the last push except the removal of an internal coupling (the FIX-profile alignment no longer writes the statistics' expected-cycle value; it is derived from the consumer state).

herisson-88:
Thanks for the review — it was right on most points, and it triggered a second, adversarial pass of my own that found more. Everything is now in ten self-contained theme commits (force-pushed; every commit the review touched ends with an "After review" paragraph; all ten build against SDK 150, the first also against SDK 149; `type: description` subjects). The description's table has the current hashes. One reply instead of the running commentary — here is the whole picture.

## Your findings

| # | Finding | Outcome |
|---|---|---|
| 1 | SDK 149 build break (`connect(cpu, rapidStart)`) | **Fixed.** `sdkConnect()` selects the one- or two-argument overload at compile time; `--rapid-start` warns and is ignored on 149. Verified by building the tree against `DirettaHostSDK_149` and `_150`. |
| 2 | Heap overflow in the EOF drain | **Not reproducible, and now provably absent.** Every copy is clamped by `samplesNeeded`, so `totalSamplesRead` never exceeds `numSamples`; the decode suite runs clean under AddressSanitizer on FLAC/ALAC/MP3 (codecs that flush several frames at EOF). An explicit guard makes the invariant visible. |
| 3 | v2.4.4 bit-depth regression | **Half right, fixed.** 24-bit sources were already 24 → 32 → 16 (32 only if the sink refuses 24, TEAC-safe); 16-bit sources wrongly offered 32 first — back to 24 → 16, 32 last. |
| 4 | Prefetch thread not masking shutdown signals | **Fixed** in `demoteToHelperThread()` (prefetch and preload threads). |
| 5 | `stopThread()` hanging on a stalled read | **Fixed.** The inner HTTP context is opened with the reader's own interrupt callback (abort on stop, otherwise chained to the decoder's stall deadline). |
| 6 | Stats data race | **Fixed.** Relaxed atomics (worker writes, SIGUSR1 reads). |
| 7 | `SystemCallFilter=` removed | **You're right, restored.** Tens of nanoseconds per syscall is not worth dropping hardening. |
| — | Port wait uninterruptible | **Fixed, and made opt-in.** `--port-strict` / `PORT_STRICT=1` waits for the configured port, checking the stop signal every 100 ms, messages on stderr. The default keeps libupnp's behaviour (port+1, announced over SSDP, which most control points follow) plus a warning naming the port actually used — up to a minute without a renderer after every hot restart is too high a price for a default. Root cause remains libupnp without `SO_REUSEADDR`; the wait is the workaround that works with the distro package. |
| — | Doubled timestamps, buffers "raised" 1.0 → 3.0 s, Qobuz-only probe cap | Not from this branch: `TimestampedStreambuf` and the `[Nms]` drain prefix predate it; `PCM_REMOTE_BUFFER_SECONDS = 3.0` has been in the code since `6cdabd5` (the docs said 1.0 and were corrected); the probe cap applies to every URL. |

## What the second pass found (all folded in)

- **The "stable heap" was the opposite.** `mallopt(M_MMAP_THRESHOLD, 1 GB)` is silently rejected by glibc (max 32 MB) and the following calls froze the dynamic threshold at 128 KB; meanwhile `M_TOP_PAD = 64 MB` — per arena, locked and populated under `MCL_FUTURE` — took the locked RSS from ~100 MB (2.5.15, 23 h run) to 600 MB–1 GB. Measured on the host: 618 MB → **101 MB** idle after the fix (32 MB, return checked; 1 MB pad).
- **`PrefetchReader` would have broken low-bitrate web radio**: a full 256 KB `avio_read()` before the first byte reached the demuxer = 16 s at 128 kbit/s, i.e. the v2.5.11 stall deadline fires. `avio_read_partial()` now. A preloaded next track pulled its whole 4 MB ring while the current track played — the concurrent-read corruption `e4c4428` exists for — capped at 256 KB until the track becomes current. The reader could outlive a freed `AVFormatContext` on every `open()` failure after the input was opened (dangling interrupt callback) — those exits go through `close()`. Copy outside the mutex, low-water refill, EOF/error classification, failed Range seek handled.
- **Sink negotiation failure left a PLAYING zombie** (bare `return false`): it now tears down like a fatal decode error, so the control point sees STOPPED and the idle release runs.
- **`setSink()` default had changed silently**: 2.5.15 passed the computed cycle time, not 0. The default is 2.5.15's value again; `--sink-buffer-ms 0` is the explicit request for the sink's own default.
- Smaller: RT mailbox shared a cache line with what the worker writes every call and was `exchange`d every decode iteration (own line, relaxed poll); seek queued while paused survived a URI change; DoP even-frame invariant in FIX alignment; warnings on unbuffered stderr were one `write(2)` per character; the `--quiet` null sink was a static destroyed before `ios_base::Init`; the hysteresis comment overstated the wake-up reduction (now sized to reach LOW, ~20/s); doc contradictions (`--cycle-time` examples off by ~3×, README help block, SystemCallFilter sentence); four ring unit tests added (S24 hint across `clear()`/`resize()`, timeout default, whole-frame pushes, 32→16); `tests/bitexact_suite.sh` makes the bit-exact claim reproducible against generated WAV/FLAC.

Two of my own earlier claims were withdrawn after measuring 2.5.15 with the same harness: the EOF drain is hygiene (identical frame counts and hashes on FLAC/ALAC/MP3), and the partial-frame push was latent (the ring never got that full under the 50 % throttle).

## Listening

On this chain (Holo Audio Red, JPLAY, Qobuz 16/44 → 24/96 and local FLAC) the branch defaults are clearly preferred to 2.5.15. None of the three SDK options survived a longer listen: `auto-sdk` (FIX 2 ms cycles, working since the alignment fix) sounded worse than VARIABLE, `rapid-start` had no measurable or lasting effect, `sink-buffer-ms 100` was dropped. They are exposed and testable, not recommended, and off by default.

DSD was tested even though the branch has no DSD-specific code change: native DSD64 and DSD128 from DSF, and DSD64 from DFF, play on the Red (0 underruns over 45 s each, cadence as expected); the decoder suite covers the same files with and without the prefetch thread and with seek sequences (the DFF being a bit-reversed copy of the DSF, both decode to the same bytes). DoP on the Red → Holo Spring 2 KTE: `DOP=1` on a DSF gives music plus constant hiss, `DOP=msb` plays identically to native DSD — the conclusion issue #80 reached; behaviour untouched, noted in the description.

## Known and left for you

- `src/test_audio_memory.cpp`: the two DoP tests fail on v2.5.15 as well (they predate the v2.5.10 byte-pair fix); not touched here.
- libupnp `--enable-reuseaddr` at build level would remove the port wait entirely.
- The tuner drop-in change (no process-wide FIFO 90) is untested on targets other than the Red.

Take it commit by commit or as a whole — each one reverts cleanly on its own.

PR #91, opened 2026-09-08
fix: seek — flush the Diretta ring, never drop a request, allow it while paused

herisson-88:
Split out of #90 (kept open as the informational, whole-picture PR: analysis documents, review thread, listening notes). Each PR here is rebased on `main` and carries its own CHANGELOG lines. Independent of the other PRs.

**What**: (1) the Diretta ring is flushed after the decoder seeks (`DirettaSync::flushForSeek()`, prefill restarts) — before, everything already buffered from the old position kept playing first, so a seek was heard seconds late and repeated seeks piled up; (2) the async seek request is consumed with an `exchange()` before the target is read (the load-then-clear order could drop a seek); (3) a seek while paused is queued and applied on resume, cleared on Stop or on a URI change; (4) `Seek` with a non-time unit is ignored instead of being applied as seconds.

**Verified**: builds against SDK 150; decoder-level seeking checked with `test_decode --seeks` (5-seek sequences and 30 random rapid seeks on FLAC 16/44, FLAC 24/192, ALAC, MP3 — the harness is in the prefetch/tests PR); seeks from JPLAY on a Holo Audio Red.

You already called this one solid in the #90 review.

**Review of the split PR (fixed here):** on `main`, `DirettaRingBuffer::clear()` forgets the S24 alignment hint and only `open()` re-sets it, so a seek on a 24-bit track fell back to sample sniffing — the three same-format ring restarts (seek, resume, quick resume) now share `resetRingForRestart()`, which keeps the hint (#95 makes `clear()` itself keep it; this PR is correct without it). `Seek TRACK_NR/TRACK_INDEX=1` restarts the track (it was applied as "1 second" before, then ignored). The seek flag is polled with a plain load before the `exchange()`. Not changed, for lack of evidence: the DSD quick-resume path sends a short silence before clearing the ring and a seek does not — no click was heard; say if you know one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+- **Seek**: (1) the Diretta ring is now flushed after the decoder seeks (`DirettaSync::flushForSeek()`, prefill restarts) — before, everything already buffered from the old position kept playing first (up to 0.5 s local / 3 s remote ring), so a seek was heard seconds late and repeated seeks piled up while the reported position had already jumped; (2) the async seek request is consumed with an `exchange()` before the target is read — the old load-then-clear order could drop a seek that arrived in between (scrubbing control points send several per second); (3) a seek while paused is queued and applied on resume instead of being refused ("Cannot seek when not playing"), and cleared on Stop or on a URI change; (4) `Seek` with a non-time unit (`TRACK_NR`) is ignored instead of being applied as seconds. Decoder-level seeking verified with `test_decode --seeks` (harness in the prefetch/tests PR) (5-seek sequences and 30 random rapid seeks on FLAC 16/44, FLAC 24/192, ALAC, MP3; identical output with and without the prefetch thread).
+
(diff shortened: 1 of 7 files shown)


cometdom:
Re-reviewed against the concerns raised on #90 for this commit (seek-while-paused, the lost-seek race, ring flush on resume). All three are correctly addressed:

- `seek()` now only rejects `STOPPED`, not `!= PLAYING` — a seek while `PAUSED` is queued and applied by the first `process()` after resume.
- The queued flag is now consumed with `m_seekRequested.exchange(false, ...)` *before* reading `m_seekTarget`, instead of load-target-then-clear-flag — the old order could lose a seek that arrived in the small window between the two operations (scrubbing control points send several requests per second).
- `m_seekRequested` is explicitly cleared on both forced-reopen/URI-change and `stop()`, so a seek queued while paused can no longer apply against a different track.
- New `setSeekCallback()` → `DirettaSync::flushForSeek()` drops whatever the ring/output already buffered from the old position at the moment the seek actually lands.

No issues found. Good to merge as far as this review is concerned.

PR #92, opened 2026-09-08
chore: tuner drop-ins without process-wide FIFO, doc and conf corrections

herisson-88:
Split out of #90 (kept open as the informational, whole-picture PR: analysis documents, review thread, listening notes). Each PR here is rebased on `main` and carries its own CHANGELOG lines. Independent of the other PRs — no code.

**What**: the tuner drop-ins no longer set `CPUSchedulingPolicy=fifo` / `Priority=90` on the whole process (every control thread ran above the audio worker at `RT_PRIORITY`, and `NICE_LEVEL` was a no-op — the renderer sets SCHED_FIFO per thread itself); `docs/CONFIGURATION.md`: bit 8 of `--thread-mode` is not `SOCKETNOBLOCK` (commented out in the SDK header), `FEEDBACKOFFSET` is a 3-bit field, `--cycle-time` default/range and examples recomputed from `DirettaCycleCalculator` (they were off by ~3×); conf + README: the remote buffer defaults are 3.0 s / 500 ms (what the code has had since `6cdabd5`), not 1.0 s / 150 ms; a comment on the cost of `SystemCallFilter=` (tens of nanoseconds per syscall — an A/B knob, not something to drop).

**Verified**: the drop-in change on a Holo Audio Red host (x86 Fedora RT); untested on other targets.

**Review of the split PR (fixed here):** the `--cycle-time` range statement moved to #94, which is the PR that widens the check (on `main` the binary warns outside 333-10000); the adaptive-buffer table in `docs/CONFIGURATION.md`, the conf's thread-mode and `CYCLE_TIME` comments and the web UI descriptions now say the same numbers as the code; the drop-in comment no longer names a thread that does not exist on `main`, says that the decode thread is SCHED_FIFO only when `CPU_DECODE` is set (without it, it is SCHED_OTHER with the new drop-in — a change for tuner users, now documented), and that the tuner must be re-run on an existing host; "worker at 80" → `RT_PRIORITY` (default 50); the `SystemCallFilter=` comment no longer claims a per-syscall cost on kernels that cache the verdict.


cometdom:
Checked the one item this review flagged against #90's version of the systemd unit: whether `SystemCallFilter=` had been weakened/removed. It hasn't — `main` already carries `SystemCallFilter=~@mount @keyring @debug @module @swap @reboot @obsolete` unchanged, and this PR only adds a clarifying comment about its per-syscall cost ("comment it out only for an A/B, not as a default"). Since this PR is rebased on current `main` rather than on #90's branch, the regression I'd flagged there simply isn't present here.

No issues found on the rest of the diff (tuner conf/doc corrections). Good to merge.

PR #93, opened 2026-09-08
fix: --port-strict — wait for the configured UPnP port after a hot restart (opt-in)

herisson-88:
Split out of #90 (kept open as the informational, whole-picture PR: analysis documents, review thread, listening notes). Each PR here is rebased on `main` and carries its own CHANGELOG lines. Independent of the other PRs.

**What**: libupnp is built without `SO_REUSEADDR` on its miniserver; after a hot restart, with the control point's connections still in `TIME_WAIT`, `bind()` fails and libupnp silently takes port+1. It announces the new LOCATION over SSDP and most control points follow; JPLAY keeps its cached address and never reaches the new instance (seen twice in one session, 4005 → 4006). `--port-strict` / `PORT_STRICT=1` makes `UPnPDevice::start()` retry every 2 s, up to 75 s, checking the stop signal every 100 ms, messages on stderr so they show under `--quiet`. **Default unchanged** (libupnp's behaviour) plus a warning naming the port actually used — a minute without a renderer after every hot restart is too high a price for a default. The root fix is libupnp's `--enable-reuseaddr` at package/build level.

**Verified**: builds against SDK 150; on the Red host: "Port 4005 busy … libupnp offered 4006" → "Port 4005 acquired after 54s".

**Review of the split PR (fixed here):** the first version retried `UpnpInit2()`/`UpnpFinish()` every 2 s while holding `UPnPDevice`'s state mutex; on `main` the signal handler calls `exit()` on that same thread, and `~UPnPDevice()` → `stop()` re-takes the mutex — a SIGTERM during the wait deadlocked until systemd's SIGKILL. It also logged "acquired after 76 s" after giving up. Now: a bare `bind()` probe (no `SO_REUSEADDR`, fails with `EADDRINUSE` under exactly the TIME_WAIT condition libupnp hits, leaves no TIME_WAIT of its own) polled every 100 ms for up to 75 s, before libupnp is initialised and outside the mutex; one `UpnpInit2()`; the port libupnp took is the verdict. `--port` is range-checked. `PORT_STRICT` is in `install.sh`'s known keys (an upgrade dropped it as obsolete). The messages use `LOG_WARN` (stdout on `main`; visible under `--quiet`, which lowers the level) — the earlier "stderr" claim was wrong. Verified on the host, against the real TIME_WAIT left by the instance just stopped (`ss -tan` showed 7 sockets):

```
[12:27:34.361] [WARN] [UPnPDevice] Port 4005 busy (TIME_WAIT from the previous instance?) — waiting for it, up to 75 s
[12:27:35.861] [WARN] [UPnPDevice] Port 4005 free after 1 s
[12:27:35.881] ✓ UPnP initialized on 192.168.1.24:4005
```

and, SIGTERM sent while the wait was in progress (no "free after" line yet): `exited rc=0, 85 ms after SIGTERM` — the first version deadlocked here.


cometdom:
Checked specifically for the concern raised on #90: since `--port-strict` reopens up to a 75s wait for the fixed port to free up (TIME_WAIT), does it reopen the v2.5.13 Ctrl-C-during-startup-retry regression?

It doesn't. The wait loop sleeps in 100ms increments and checks `m_stopSignal` on every iteration:

```cpp
for (int slept = 0; slept < PORT_RETRY_INTERVAL_S * 10; slept++) {
    if (m_stopSignal && !m_stopSignal->load(std::memory_order_acquire)) {
        LOG_WARN("[UPnPDevice] Startup cancelled while waiting for port " << m_config.port);
        return false;
    }
    std::this_thread::sleep_for(std::chrono::milliseconds(100));
}
```

`DirettaRenderer::start()` wires `m_upnp->setStopSignal(stopSignal)` before this path runs, so a Ctrl-C during the wait aborts within ~100ms rather than sitting out the full retry window. The feature is also opt-in (`--port-strict`/`PORT_STRICT=1`), so default behavior is unaffected either way.

No issues found. Good to merge.

PR #94, opened 2026-09-08
feat: SDK usage — setSink buffer time, connect on the audio core, auto-sdk and rapid-start, negotiated profile log

herisson-88:
Split out of #90 (kept open as the informational, whole-picture PR: analysis documents, review thread, listening notes). Each PR here is rebased on `main` and carries its own CHANGELOG lines. Independent of the other PRs. Builds against SDK 149 and 150.

**What**: `--sink-buffer-ms` (`setSink()`'s documented sink buffer time — v2.5.15 passed the computed cycle time there, which stays the default; `0` = sink default); `Sync::connect()` is given the first `--cpu-audio` core (`sdkConnect()` selects the one- or two-argument overload at compile time, so SDK 149 still builds; `--rapid-start` warns and is ignored there); `--transfer-mode auto-sdk` (`Sync::configTransferAuto`, the SDK sample host's mode); the negotiated profile (cycle, min cycle, cycle size/packets, mode, MS mode, latency, `SinkInfo`) logged at each `OPEN`. **FIX profiles work**: when the SDK negotiates a fixed cycle (`auto-sdk` does, 2 ms on a Holo Red) it expects exactly `getCycleSize()` bytes per `getNewStream()`; the renderer's 1 ms buffers left the target silent — the callback buffer is now aligned on the negotiated cycle (`alignBufferToNegotiatedCycle()`, DoP even-frame rule kept). Docs/conf for the new options.

**Listening** (Red, JPLAY, Qobuz and local FLAC): none of the three options survived a longer listen — `auto-sdk` sounded worse than VARIABLE, `rapid-start` had no measurable or lasting effect, `sink-buffer-ms 100` was dropped. They are exposed and testable, not recommended, and off by default; the defaults reproduce v2.5.15's SDK calls.

**Review of the split PR (fixed here):** unpinned (`--cpu-audio` unset), `connect()` gets 0 as v2.5.15 did — the defaults now reproduce v2.5.15's SDK calls exactly (the first version passed -1). The `--cycle-time` accepted range change (333-10000 → 100-50000, the auto value being outside the old range) was undocumented: CHANGELOG, docs, conf and web UI say so now. `--cycle-time` is honoured as the target cycle in `auto-sdk` (it was silently ignored). `alignBufferToNegotiatedCycle()` applies to every FIX profile, `fixauto` included — documented; whether `fixauto` played with 1 ms buffers on v2.5.15 was not tested; it warns when `getCycleSize()` is not a whole number of frames (24-bit, DoP even-frame rule). `SINK_BUFFER_MS`/`RAPID_START` in the web UI and `install.sh`'s known keys. `DIRETTA::Profile::FIX` instead of a magic 1; comments and the commit message state the right numbers (the message contradicted itself on the `--sink-buffer-ms` default). Not changed: the alignment is run before `connect()` and after `connectWait()`; if the SDK ever changed the mode between the two, the second call could not restore the VARIABLE values — the profile is logged at both points, so that would be visible; not observed.


cometdom:
Verified the critical item from #90 empirically, not just by reading the diff: the original `connect(sdkConnectCpu, m_config.rapidStart)` call broke the build against SDK v149, which only declares `bool connect(int)`. I confirmed the break by actually compiling `main` against `DirettaHostSDK_149` (`error: no matching function for call to 'DirettaSync::connect(int&, bool&)'`).

This PR's fix compiles cleanly against both SDKs. I checked this branch out and built it against `DirettaHostSDK_149` directly (`DIRETTA_SDK_PATH=... make`) — succeeds ("Build complete: bin/DirettaRendererUPnP") — in addition to the default-detected v150 tree. The SFINAE/`if constexpr` overload resolution is the right approach here (resolves the overload at compile time instead of pinning to one SDK version):

```cpp
template <typename S, typename = void>
struct SdkHasRapidStart : std::false_type {};
template <typename S>
struct SdkHasRapidStart<S, std::void_t<decltype(std::declval<S&>().connect(int{}, bool{}))>>
    : std::true_type {};

template <typename S>
static bool sdkConnect(S& sync, int cpu, bool rapidStart) {
    if constexpr (SdkHasRapidStart<S>::value) {
        return sync.connect(cpu, rapidStart);
    } else {
        (void)rapidStart;
        return sync.connect(cpu);
    }
}
```

No issues found on the rest of the diff (`--sink-buffer-ms`, `auto-sdk`, `--rapid-start`, negotiated-profile logging, FIX-profile alignment). Good to merge — this one's the base a few of the others build on, so I'd merge it first.

PR #95, opened 2026-09-08
fix: S24 hint across pause/resume, whole-frame ring pushes, RT-thread logging, EOF drain, sink negotiation, journald flush

herisson-88:
Split out of #90 (kept open as the informational, whole-picture PR: analysis documents, review thread, listening notes). Each PR here is rebased on `main` and carries its own CHANGELOG lines. Independent of the other PRs (#96 is stacked on this one). Builds against SDK 149 and 150.

**What** (details in the CHANGELOG lines): the S24 alignment hint survives `clear()` (pause → resume while the music is quiet locked the ring to LSB on x86: full-scale noise), timeout default MSB, no `#if __aarch64__` special case; PCM pushes are whole-frame (latent); `configureSinkPCM()` returns `false` instead of throwing (orders 32→24→16 / 24→32→16 / 24→16→32 — 32-bit never first for a 16/24-bit source, v2.4.4 kept), a 24/32-bit source on a 16-bit-only sink is truncated explicitly (`push32To16`); sink refusal tears the track down instead of leaving a PLAYING zombie; `getNewStream()` no longer touches iostreams (atomic mailbox logged by the decode thread); FFmpeg's EOF drain sequence (output-neutral on FLAC/ALAC/MP3, measured — kept as correctness-by-contract); `TimestampedStreambuf::sync()` so log lines reach journald.

**Verified**: ring unit tests (four added: S24 hint across `clear()`/`resize()`, timeout default, whole-frame pushes, 32→16 — `make test` reports 25 passed, 2 failed: the two DoP tests that fail on `main` as well, they still assert the pre-233b520 byte order); decode paths bit-exact against v2.5.15 and against source PCM; listened to on a Holo Audio Red (PCM, native DSD, DoP).

**Review of the split PR (fixed here):** dropping the `#if __aarch64__` MSB force regressed one input class on ARM: the alignment hint is only produced for 24-bit sources, so a 32-bit source on a sink that refuses 32-bit (packed to 24 by `configureRingPCM()`) fell back to sample sniffing, which picks LSB on real 32-bit data — noise, where the ARM force played the top 24 bits (x86 already did this on v2.5.15). `configureRingPCM()` now sets the MSB hint itself whenever a 32-bit container is packed to 24 bits (FFmpeg only ever left-justifies). `consumeRtEvents()` loads before it exchanges, and the mailbox is drained from the idle branch of the audio thread too (an underrun raised during the end-of-track drain was logged at the start of the next track). The `!continuePlayback` teardown is skipped when the state is no longer PLAYING: `DirettaRenderer::stop()` moves the engine to STOPPED before it raises the flag the callback bails on, so that case is a Stop in flight, not a refused track (v2.5.15 did not tear down there either). A 24-bit source upgraded to a 32-bit sink is logged. `test_s24_hint_survives_clear` now fails on the old `clear()` (it passed on v2.5.15 as written). `docs/CLAUDE.md`'s Bit Depth Handling and S24 Alignment sections describe this PR's rules (they contradicted it). CHANGELOG: `configureSinkDSD()` still throws; the pause/resume condition is 48000 samples, not "a second".


cometdom:
Verified both critical items flagged on #90 against this PR's diff.

**EOF-drain `size_t` underflow (heap overflow risk) — fixed.** The original bug: the drain loop deliberately keeps pulling decoder frames past `totalSamplesRead == numSamples` to flush codec lookahead (ALAC/AAC/MP3/Vorbis/WavPack); once `totalSamplesRead > numSamples`, `size_t samplesNeeded = numSamples - totalSamplesRead;` underflowed to a huge value, defeating the `std::min(frameSamples, samplesNeeded)` clamps at the copy sites and letting `memcpy_audio()` write past the end of the exactly-sized output buffer. Fix:

```cpp
size_t samplesNeeded = (totalSamplesRead < numSamples) ? numSamples - totalSamplesRead : 0;
```

Clamps to 0 once the caller's buffer is full; excess decoded samples still route to the FIFO for later consumption, so the intended drain behavior is preserved.

**Bit-depth regression (v2.4.4) — fixed, and improved beyond the pre-#90 baseline.** `configureSinkPCM()` now returns `bool` instead of `throw`ing `std::runtime_error` on total format-negotiation failure — callers check the return and fail the track cleanly instead of crashing the process. The fallback order is now genuinely ordered per source depth instead of a flat 32→24→16:

```cpp
if (inputBits >= 32)      { order = {32, 24, 16}; }
else if (inputBits == 24) { order = {24, 32, 16}; }
else                      { order = {24, 16, 32}; }
```

32-bit is only ever tried as a last resort for 16/24-bit sources (after the preferred depth *and* the other non-32 depth have failed), satisfying CLAUDE.md's "never prefer 32-bit for a 16/24-bit source" invariant while adding a safety-net fallback that the pre-#90 code didn't have.

No issues found on the rest of the diff (S24 hint, whole-frame pushes, RT-thread logging, journald flush). Good to merge.

PR #96, opened 2026-09-08
perf: quieter host, HTTP prefetch thread, FLAC bypass, decode test harness and suites

herisson-88:
Split out of #90 (kept open as the informational, whole-picture PR: analysis documents, review thread, listening notes). Each PR here is rebased on `main` and carries its own CHANGELOG lines. **Stacked on #95** (its commit shows here until that one is merged): the hysteresis and self-pipe changes touch the same lines. Builds against SDK 150.

**What**: decode-thread hysteresis (fill to 70 %, sleep until 30 % — ~20 wake-ups/s instead of a 10 ms poll); real `--quiet` (stdout discarded, warnings on stderr); stable heap (`mallopt` 32 MB / -1 / 1 MB before `mlockall` — measured 101 MB locked RSS idle); self-pipe shutdown (the handler called `exit()` on a live condvar waiter: 45 s hang + SIGABRT); dead 1 Hz threads removed, preload thread demoted, FFmpeg probe capped; cadence stats in `dumpStats()`. `PrefetchReader`: every HTTP source read by a `SCHED_OTHER` thread on the `--cpu-other` cores with `avio_read_partial()`, 4 MB ahead of the demuxer through a custom `AVIOContext`, own interrupt callback (stop never waits on a stalled read), preloaded next track capped at 256 KB until current (e4c4428), shutdown signals masked; `--no-prefetch` for A/B. FLAC takes the bit-perfect bypass (the decoder's output format decides, ALAC/WavPack stay on the resampler). Dead `src/sync/` copy removed. Test tooling: `make test-decode` (`--seeks`), `tools/range_server.py`, `tools/make_test_audio.py`, `tools/dsf2dff.py`, `tests/decode_suite.sh`, `tests/bitexact_suite.sh`.

**Verified**: `tests/bitexact_suite.sh` 8/8 (synthetic WAV/FLAC hashed against source samples, prefetch vs not); `tests/decode_suite.sh` 20/20 on real FLAC 16/44, FLAC 24/192, ALAC 24/96, MP3 (full decode, 5-seek and 30-random-seek sequences, identical hashes with and without the prefetch thread); DSD64/DSD128 DSF and DSD64 DFF through the same suite; ASan clean; listened to on a Holo Audio Red — the branch defaults are clearly preferred to v2.5.15 on that chain.

**Review of the split PR (fixed here):** the reader kept a pointer to the decoder's `AVIOInterruptCB` inside the `AVFormatContext`; on an `avformat_open_input()` failure FFmpeg frees that context before the reader thread stops — the callback is copied by value now. The custom `AVIOContext` mirrors the inner context's `seekable` flag instead of "size known" (FFmpeg's http marks a response without `Accept-Ranges`/`Content-Range` as streamed). The URL is passed to `avformat_open_input()` for probing (`s->url` and the extension score were empty). The first-packets hex dumps in the DSD read path were `printf()` and escaped `--quiet` as orphan lines; DEBUG only. The probe comment claimed header-only analysis. `tests/decode_suite.sh`'s error filter dropped the summary line whenever `readTimeout=0`. `.PHONY`, `NO_PREFETCH` in `install.sh`'s known keys, commit wording (5-100 ms; the decode thread no longer *blocks* in network reads — open and Range seeks still run on the calling thread, DFF is not prefetched). **Measured, not changed:** a Range seek the server refuses ends the track with the prefetch thread — and without it too, on this FFmpeg 8.1.2 (both modes stop after the refused seek); the MIME hint FFmpeg's own http would have offered the probe is not forwarded — FFmpeg 8.1.2 fails `audio/L16` on its own path as well, so a fix could not be verified here. Not changed: with `--cpu-decode` set and `--cpu-other` unset the prefetch/preload threads inherit the decode core's affinity (SCHED_OTHER there); three near-identical affinity/sigmask helpers remain.


cometdom:
Went through every item flagged on #90 that touches this PR's scope. Summary: the two safety-relevant findings are both genuinely fixed; a couple of minor/cosmetic items remain, none blocking.

**Stats data race — fixed.** All the counters written by `getNewStream()` (the SDK's real-time worker thread) and read by `dumpStats()` (from a different thread via the `SIGUSR1` handler) are now `std::atomic<int64_t>`: `m_streamLastNs`, `m_streamIntervalMinNs/MaxNs/SumNs/Count/Late`.

**`PrefetchReader::stopThread()` indefinite-hang risk — fixed, verified by tracing the actual interrupt wiring, not just presence of a callback.** `stopThread()` itself still does a bare, untimed `m_thread.join()` — so I traced whether the join is actually bounded by something else. It is:

- The inner HTTP `AVIOContext` (the one `avio_read_partial()` blocks on) is opened with `PrefetchReader`'s *own* interrupt callback (`m_ownInterrupt`), which checks `m_stopRequested` **unconditionally first**, before chaining to the decoder's stall-deadline callback:
  ```cpp
  int PrefetchReader::interruptCb(void* opaque) {
      auto* self = static_cast<PrefetchReader*>(opaque);
      if (self->m_stopRequested.load(std::memory_order_acquire)) return 1;
      if (self->m_interruptCb && self->m_interruptCb->callback) {
          return self->m_interruptCb->callback(self->m_interruptCb->opaque);
      }
      return 0;
  }
  ```
- `stopThread()` sets `m_stopRequested = true` *before* joining. FFmpeg's tcp/http protocol polls its interrupt callback roughly every 100ms while blocked, so a blocked `avio_read_partial()` aborts promptly on the next poll regardless of any deadline state — the join is bounded independent of `m_readDeadlineNs`.
- Separately, I confirmed the shared-deadline chain also works correctly for actual stall *detection* (not just shutdown): `m_readDeadlineNs` is armed for the full duration of the decode thread's `av_read_frame()` call, including when that call blocks indirectly inside `PrefetchReader::readFromRing()` waiting for the prefetch ring to fill — so a real network stall while `PrefetchReader` is the one blocked still trips the existing 5s stall timeout and triggers the existing reconnect path, which then calls `stop()`/`stopThread()` and unblocks via the `m_stopRequested` path above.

So the shared `AVIOInterruptCB` plumbing isn't just a well-intentioned attempt — it closes the gap via two independent, correctly-composed mechanisms (unconditional stop flag for shutdown, chained deadline for live-stall detection). No further action needed here.

**Minor items still present, non-blocking:**
- `AudioEngine::demoteToHelperThread()` ends with a dead `(void)name;` — `name` is already used a few lines above in the `DEBUG_LOG` call. Cosmetic.
- `demoteToHelperThread()` re-implements the `pthread_sigmask` SIGINT/SIGTERM-blocking inline rather than calling the existing `blockShutdownSignalsOnThisThread()` helper from `main.cpp`/`DirettaRenderer.cpp` — functionally correct (verified: it does block both signals as the first thing in the thread), just a third near-duplicate of the pattern instead of a shared call.
- `PrefetchReader` uses its own mutex+condvar byte ring rather than reusing `DirettaRingBuffer`'s lock-free SPSC design — a deliberate, reasonable choice given it's not on the hot path (a plain `SCHED_OTHER` thread on `--cpu-other`), just noting the divergence from the project's usual pattern.

I didn't re-flag the smaller/secondary items from the original #90 review (`canBypass()` compressed-format guard, hardcoded Qobuz/Tidal provider match, buffer-constant increases without re-measurement, etc.) — nothing new turned up there and they're not safety-relevant.

Good to merge (after #95, which it depends on).

PR #97, opened 2026-09-09
fix: build on SDK 149 — is_MSmode() only exists in SDK 150

herisson-88:
## Problem

Building v2.5.17 against `DirettaHostSDK_149` fails (reported on Fedora via `fedora-audiophile-setup` option 12, clang/`LLVM=1` build):

```
src/DirettaSync.cpp:2405:48: error: use of undeclared identifier 'is_MSmode'
2405 |              << " msMode=" << static_cast<int>(is_MSmode())
```

`DirettaSync::logNegotiatedProfile()` (added in PR #94) calls `Sync::is_MSmode()` unconditionally, but that accessor only exists in SDK 150. Every other getter in that log line (`getCyclePackets`, `getMinCycleTime`, `getLatency`, `getSinkInfo`) is available in SDK 149, so this is the single blocker. The CLAUDE.md contract is that SDK 149 keeps building.

## Fix

Resolve the accessor at compile time with the same SFINAE pattern already used for `connect(int, bool)` (`SdkHasRapidStart`):

- new `SdkHasIsMSmode<S>` trait + `sdkMSmodeString()` helper next to `sdkConnect()`
- `logNegotiatedProfile()` logs the real value on SDK 150 and `n/a` on SDK 149

No behaviour change on the audio path; only the diagnostic log line is affected.

## Verification

- The SDK is not available in the CI environment where this was written, so the full file was not compiled there. The SFINAE helper was compiled and run in isolation against two mock classes (one with `is_MSmode()`, one without) with both `g++` and `clang++ -std=c++17 -Wall -Wextra`: prints `n/a` and `1` respectively.
- Please confirm on a real SDK 149 host with `sudo ./setup.sh --only install-drup` (or `make LLVM=1`).

The 18 remaining warnings in the reporter's output are unrelated: the `-Wignored-qualifiers` ones come from the SDK's own `Receive.hpp` under clang, the rest are pre-existing unused variables.

https://claude.ai/code/session_017n6naEwNTsJwEEqXcFYrZk

---
_Generated by [Claude Code](https://claude.ai/code/session_017n6naEwNTsJwEEqXcFYrZk)_

### DIFF
--- a/src/DirettaSync.cpp
+++ b/src/DirettaSync.cpp
@@ -99,6 +99,25 @@ static bool sdkConnect(S& sync, int cpu, bool rapidStart) {
     }
 }
 
+// SDK 150 also added Sync::is_MSmode(); SDK 149 has no such accessor.
+// Same compile-time resolution: log the negotiated MS mode when the SDK
+// exposes it, "n/a" otherwise, instead of failing the build on SDK 149.
+template <typename S, typename = void>
+struct SdkHasIsMSmode : std::false_type {};
+template <typename S>
+struct SdkHasIsMSmode<S, std::void_t<decltype(std::declval<S&>().is_MSmode())>>
+    : std::true_type {};
+
+template <typename S>
+static std::string sdkMSmodeString(S& sync) {
+    if constexpr (SdkHasIsMSmode<S>::value) {
+        return std::to_string(static_cast<int>(sync.is_MSmode()));
+    } else {
+        (void)sync;
+        return "n/a";
+    }
+}
+
 class RingAccessGuard {
 public:
     RingAccessGuard(std::atomic<int>& users, const std::atomic<bool>& reconfiguring)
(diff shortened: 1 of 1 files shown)


cometdom:
Thanks for catching this independently — nice confirmation the diagnosis was right (SDK 149 isn't one fixed snapshot; `is_MSmode()` is present in some sub-revisions and not others, same class of gap `connect()` already had).

This one's already fixed and released: **v2.5.18** (commit 8da1980), same SFINAE approach — `SdkHasMSmode<S>`/`sdkMsMode()` next to `sdkConnect()`, logging a sentinel instead of failing the build when the SDK doesn't expose the getter. Verified against both a local SDK 149 and 150 build, plus an isolated standalone SFINAE test against mock classes with/without `is_MSmode()`.

One small difference worth noting: this PR logs `"n/a"` (string) for the missing case, where v2.5.18 logs `-1` (int) — `"n/a"` reads better in the log than a bare `-1` that could be misread as an actual mode value. Small enough that I'll fold it into the next release rather than asking you to rebase this PR.

Closing as already shipped in v2.5.18 — thanks again for the fast turnaround on this.

cometdom:
Merci d'avoir sauté sur ce rapport aussi vite, @herisson-88 — c'est très appréciable, et ça confirme au passage que le correctif publié dans v2.5.18 était le bon. On garde le `"n/a"` de ta version pour le prochain correctif cosmétique.

PR #98, opened 2026-09-20
fix: clicks on Stop, Pause, track skip and seek — 10 ms PCM fades, ring dropped behind the shutdown silence

herisson-88:
## The report

An Audirvana user on Audiophile Style reported small "scritch" clicks with the renderer, regularly but not always, on **Stop** and on **track skip**. Reproduced several times with Audirvana → DRUP (v2.5.16 build) → Holo Audio Red, PCM; confirmed fixed by listening with this branch on the same chain.

Looking for the same defect elsewhere turned up the other places where PCM is cut or restarted in the middle of the music (seek, resume from pause, rebuffering). They are fixed here too, with the same ramp and the same code.

## Root cause

`getNewStream()` goes from a music buffer straight to a zero buffer as soon as a shutdown silence is requested:

```cpp
// Shutdown silence
int silenceRemaining = m_silenceBuffersRemaining.load(std::memory_order_acquire);
if (silenceRemaining > 0) {
    fillSilence(dest, currentBytesPerBuffer);   // previous buffer was music at full level
```

That is a step in the waveform. How loud the click is depends on the sample value where the cut lands — nothing on a quiet passage, a clear click on a loud one, which matches "regularly, not always".

Why Audirvana makes it so easy to hear: it does not skip with a bare `SetAVTransportURI`. Captured on the wire (port 4005), three skips in a row, always the same sequence:

```
16:29:52.430  Stop
16:29:52.455  SetAVTransportURI   CurrentURI = …/audirvana/audio_f_85.wav
16:29:52.455  Play
16:29:53.351  SetNextAVTransportURI …
```

So every skip goes through `onStop` → `stopPlayback(false)` → `requestShutdownSilence(20)`, i.e. through the hard cut above. Pause takes the same route (`requestShutdownSilence(10)`).

Ruled out: leftovers of the previous track replayed on the `Play` that follows 25 ms later (the quick-resume path does clear the ring), and the link (no underrun logged during the skips).

The same step exists, in both directions, wherever the ring is dropped or refilled mid-track:

| Where | Music → silence | Silence → music |
|---|---|---|
| Stop, Pause, skip, format change, close | hard cut | — (next track starts from 0:00) |
| Seek (`flushForSeek()`) | ring cleared mid-music | restarts mid-waveform after the prefill |
| Resume from pause | — | restarts mid-waveform |
| Underrun → rebuffering | ring empty, nothing left to fade | restarts mid-waveform |

## The fix

**`src/PcmFade.h`** (new, header-only, no SDK dependency): 10 ms **smoothstep** ramps — no corner at either end. Q16 integer gain computed as the exact rational `r²(3N−2r)/N³`, so it is monotonic at every sample rate; samples rounded to nearest; 16/24/32-bit. No allocation, no I/O, no float: fine in the SDK callback thread. Cost: one 64-bit division per frame, during 10 ms, once per event.

**One request convention.** Two atomics, `m_fadeOutRequest` and `m_fadeInRequest`, carry a flag (`FADE_ARM` / `FADE_CANCEL`). The callback thread picks them up and computes the ramp length itself from the format it is playing; the ramp positions are plain members it owns. The work is in two out-of-line functions, `fadeOutBuffer()` and `fadeInBuffer()`, sharing one layout check (`pcmFadeLayout()`).

- **Fade-out before the shutdown silence.** `requestShutdownSilence()` arms it (before it sets the silence count); `getNewStream()` keeps popping the ring through the ramp and only then sends the silence buffers it was already sending.
- **The ring is dropped behind that silence.** The four copies of the "wait for the silence to be played out" loop (Stop, Pause, close, DSD quick resume) become one function, `playOutShutdownSilence()`, which also empties the ring as soon as the worker is on silence. Until now the ring stayed full of music after a Stop: once the silence count had run out, any callback arriving before `stop()` took effect would pop it again, at full level. Nothing is lost — `open()` and `resumePlayback()` threw that content away anyway.
- **Seek.** `flushForSeek()` calls the same `playOutShutdownSilence()` (5 buffers) before its existing reset, so the old position is faded out instead of cut dead, then arms the fade-in for the new one. It runs in the producer thread, so nothing is pushed meanwhile. DSD/DoP: unchanged, dropped at once.
- **Fade-in** after a resume from pause (armed by `resumePlayback()`) and at the end of a rebuffering (armed by the callback thread where it already raises `RT_EVENT_REBUFFER_COMPLETE`). The request is picked up with the first music buffer rather than in the prefill-silence branch: after a seek the producer refills the prefill target within one cycle, so the worker may never run that branch.
- `open()` stores `FADE_CANCEL`: a track started by `Play` is not faded in, and a ramp that a Stop interrupted half-way is dropped.

### What this does to the sound path

- Nothing upstream of the ring is touched: `sendAudio()`, `DirettaRingBuffer.h`, `AudioEngine`, `PrefetchReader`, `DirettaRenderer.cpp` — 0 lines.
- `getNewStream()`: 16 lines added, none removed, in three places — inside the existing `if (silenceRemaining > 0)` branch, inside the existing rebuffer-complete branch, and one test after the pop:

```cpp
if (m_fadeInRequest.load(std::memory_order_relaxed) != 0 || m_fadeInFramesRemaining != 0) {
    fadeInBuffer(dest, currentBytesPerBuffer);
}
```

  In steady playback both are zero (nothing writes them while the music plays), so the bytes handed to the SDK are the ones `main` hands over. Tracks started by `Play` and gapless transitions stay bit-exact; only a seek, a resume or a rebuffering is faded in, wherever it lands. The new hot members sit with `m_prefillComplete`; the cold ones are placed after the existing block so that its members keep the cache line they had.

### What keeps the previous behaviour, on purpose

- **Native DSD and DoP** cannot be scaled: `pcmFadeLayout()` refuses them, the silence starts at once as before (counted as "skipped").
- **DoP pre-encoded by the media server** (arrives as plain 24-bit PCM, `m_isDoPMode` is false): scaling it would break the 0x05/0xFA markers and the DAC would play the DSD payload as PCM for a few ms. The first buffer of either ramp is checked for alternating markers in the top byte; if found, the ramp is dropped. A false positive on real music needs two consecutive samples with top bytes exactly 0x05 then 0xFA, and only costs the old behaviour for that one event.
- **Nothing was playing** (prefill, post-online stabilisation, rebuffering): no fade-out, silence as before. If the ring runs dry (natural end of a track, where `setTrackEndCallback` drains it below 1 % before `stopPlayback(false)`), the ramp stops there.
- A buffer that is not a whole number of frames would make a ramp slide across channels, so it is refused. Every path produces whole frames today (1 ms buffers, MTU-sized buffers, the 44.1 kHz drift accumulator, `alignBufferToNegotiatedCycle()`).

### Timing

Stop and Pause take 10 ms longer; a seek blocks the decode thread ~15 ms (60 ms at most) where it took microseconds, with the lifecycle mutex held. The existing timeouts are unchanged and still cover ramp + silence: Stop/close 150 ms for ~30 ms (1 ms buffers) or ~50 ms (2 ms FIX cycle), Pause 80 ms for ~20 / ~30 ms.

## Not changed

- **`stopPlayback(true)`**: the auto-stop taken when `SetAVTransportURI` arrives *during* playback without a `Stop` first. It sends no silence at all (`sendPreTransitionSilence()` is an empty function since the format-switching issue). Audirvana never takes it and no click was reproduced there, so it is left alone rather than changed untested. A control point that skips that way still gets a hard cut.
- **A `Seek` handled between `Play` and the first decode cycle** starts mid-music without a fade-in, exactly as on `main`: `flushForSeek()` has nothing to flush yet and `open()` cancels the request. Seeking is refused while stopped, so the window is a few ms. Not observed.
- **`requestPostReconnectRebuffering()`** still cuts to silence un-faded (the return is faded in). **An underrun itself** cannot be faded out: the ring is empty.
- **Format change between two gapless tracks**: the audio callback calls `stopPlayback(false)` while the ring still holds the tail of the previous track — there is no drain like the one in `setTrackEndCallback`. With this PR that cut is faded, but the tail is still dropped (how much depends on how long the next track takes to open). Read in the code, not measured; worth a separate look.
- No dither on the ramps. 10 ms, next to digital silence.

## Verification

- **Listening**, on the commit of this PR: Audirvana → mini PC (x64, SDK 150, `NOLOG=1 LLVM=1`) → Holo Audio Red, PCM 24/44.1 and 24/48. Before: clicks reproduced on several skips and stops. After: none, on the same gestures and loud passages — skips, Stop, Stop → Play, Pause → resume. Seeks: many in a row in the middle of a loud track, sent as UPnP `Seek REL_TIME` by a second control point (the Audirvana iOS remote turned out not to send `Seek` at all — nothing on the wire), clean on both sides of the jump.
- **On the device** (`kill -USR1`, the counters added here), after that session: `Fade-outs: 30 complete, 8 skipped; fade-ins: 4`, 0 underrun. The skipped ones include DSD tracks played in the same session. On an earlier build of the branch, `fade-ins` stayed at 6 — the 6 seeks made — while nine more tracks were started with `Play`: no fade-in on a normal track start.
- **Unit tests** (`make test`, 7 new, no SDK needed): gain curve (endpoints, exactly 0.5 half-way, smooth ends, monotonic for every ramp length from 8 kHz to 768 kHz and at the clamp); all three sample widths with positive and negative full scale (starts at the music level, ends on zero, never rises, sign kept, both channels identical); rounding (±v within 1 LSB of each other all along the ramp); a ramp split over 44/45-frame buffers is byte-identical to a single pass; DoP marker detection (both phases, mono, 32-bit container, near miss, too few frames); the fade-in is the exact mirror of the fade-out and starts from zero; frames after the fade-in are bit-identical, inside the last faded buffer too.
  Result: **32 passed, 2 failed**. The 2 failures are `test_pushDSD_dop_encoding` and `test_pushDSD_dop_msb_encoding`; they fail the same way on `main` at 676a768 (25 passed, 2 failed), unrelated to this PR.
  Not covered by unit tests: the request/cancel hand-over inside `DirettaSync` needs the SDK; that part is what the on-device counters and the listening check.
- **Builds**: SDK 150 and SDK 149, `NOLOG=1 LLVM=1`. The compiler warnings are the same set as on `main` (build logs diffed).

CHANGELOG entry included under `[Unreleased]`.


cometdom:
Merged — thanks for this one. Verified before merge: full read of `PcmFade.h`'s gain math (exact smoothstep in Q16, monotonic, no overflow up to the 32768-frame clamp), clean build against both SDK 149 and SDK 150, and `make test`: 32 passed / 2 failed, the 2 failures being the same pre-existing DoP-encoding test issues already on `main` (confirmed by rebuilding main separately — 25 passed / 2 failed there, same two tests). Nice fix, and a genuinely thorough test suite to go with it.

